PR #2, opened 2026-04-25
Mimo monitor case: rounded shell + correct USB-C orientation

gregleon01:
## Summary

- **Restored the original mimo_case shell silhouette** (84×86×24.7, plan radius 18.2, face fillet 10.0) for the 3D-print-ready clamshell. The print-ready had drifted to 94×86×26 with a 6mm fillet, which read as boxy and lost the organic feel of the original.
- **Fixed a silent coordinate bug** in \`_usb_c_cutout_raw\`: build123d's \`Plane.XZ\` has its normal in -Y, so \`BuildSketch(Plane.XZ.offset(BOTTOM_Y - 0.05))\` was actually placing the cutout on the +Y face. The bug was being masked by a compensating \`rotate(Axis.Z, 180)\` in \`viewer_aligned\`. Removed both and replaced with an explicit \`Plane(origin=…, z_dir=(0,1,0))\` so the cutout lives on the BOTTOM_Y face for real.
- **Reshaped the screen recess** into a rounded rectangle (R=8) that echoes the case corners, instead of a stadium that visually clipped onto the front-face fillet.
- **Hardened validators** so dimension tweaks no longer fail with cryptic errors:
  - \`validate_assembled_bounds\` derives expected bounds from \`CASE_WIDTH/HEIGHT/TOTAL_DEPTH/BACK_PUCK_*/BUTTON_*\` instead of comparing against frozen tuple \`(84.95, 86.0, 29.6)\`.
  - \`validate_usb_alignment\` derives expectations from \`BOTTOM_Y\` and \`CASE_HEIGHT\` instead of hardcoded \`(-8.0, -40.6)\` — PCB layout changes now propagate.
- **Added a conventions header** at the top of \`mimo_case_print_ready_geometry.py\` documenting source-frame axes, where each feature should live, and the \`Plane.XZ/YZ\` normal-direction gotcha so the next contributor doesn't re-trip it.

## Why this matters

The pre-fix codebase had a coordinate convention that lied: the source frame said \"USB-C is on the bottom\" but the geometry was being placed on the top, then rotated 180° downstream to look right. This made plain-English design changes impossible — moving a feature \"up\" actually moved it down. Same trap was lurking for any feature added with \`Plane.XZ\` or \`Plane.YZ\`.

The conventions header + derived validators mean the next change like \"shrink the case to 80mm\" or \"move the screen up\" works against a source frame that matches what you see in the viewer.

## Test plan

- [x] \`models/mimo_case.step\` regenerates and \`validate_assembled_bounds\` passes
- [x] \`models/mimo_pcb.step\` regenerates and \`validate_usb_alignment\` passes
- [x] \`models/mimo_case_print_ready{,_assembled,_print_layout}.step\` all regenerate
- [x] Verified in cad-explorer viewer: USB-C cutout is on the bottom face (jump-to-bottom-view), screen recess is centered on the front face with corners that follow the case curvature, shell is properly rounded
- [ ] Reviewer: open \`mimo_case_print_ready_assembled.step\` in the viewer and rotate around — should look like a rounded near-square monitor with USB-C at the bottom edge

## Notes for reviewer

- The audit also identified a few non-blocking issues left for follow-up (constant shadowing in \`mimo_case_print_ready_geometry.py:30-46\`, magic numbers in \`PrintReadyParams\`, and an undocumented build123d axis-correction matrix in \`viewer/lib/render/glbMeshData.js\`). Listed in the working notes; happy to file follow-up issues.
- Adds \`.claude/launch.json\` so the cad-explorer dev server can be launched via Claude Preview MCP. Safe to keep or remove.


earthtojake:
i don't think you intended to push to this repo, you can fork for project-specific code!

PR #4, opened 2026-04-30
Improve CAD viewer rotation and dimensions

gnulnx:


Cem-Kaya:
I noticed the PR diff is very large because it seems to include generated model assets too, like .step, .glb, topology.bin, and topology.json files under models/erb_balance_bot/....

Not sure if that was intentional, but it might have been accidentally included along with the viewer changes. Maybe worth double checking,  removing those generated model files could make the PR much easier for the maintainer to review.

PR #6, opened 2026-05-03
fix(cad): assembly displayName uses instance.name; absolute EXPLORER_ROOT_DIR works

arronzhang:
## Summary

Two independent CAD skill bug fixes that surfaced while wiring a real downstream project (a quadruped robot whose 7 fab parts × 4 leg variants land under `exports/parts/<name>/<leg>.{step,stl}` and are then composed by `assembly/leg_module/<leg>.py` instances).

### `scripts/common/assembly_composition.py` — assembly displayName precedence

`_linked_instance_node` previously took `displayName` from the source STEP filename stem first and only fell back to `instance.name`. The generator contract makes `name` a **required**, intentionally-chosen, per-instance string; treating the filename stem as authoritative collapsed every component of a per-leg assembly to the leg label — every child in the explorer tree showed `bl` instead of `haa_shell_BL`, `thigh_BL`, etc.

This PR flips the precedence (`instance.name` → stem → instance-path component). The filename-stem fallback is retained for the path where `instance.name` happens to be empty.

The existing `test_linked_assembly_prefers_source_step_stem_over_custom_instance_name` codified the old behaviour; it is renamed to `test_linked_assembly_prefers_instance_name_over_source_step_stem` and its assertion flipped accordingly.

### `explorer/lib/cadDirectoryScanner.mjs` — absolute EXPLORER_ROOT_DIR

`normalizeExplorerRootDir()` unconditionally stripped the leading slash from `EXPLORER_ROOT_DIR` before normalization, so an absolute path like `/abs/path/exports` silently became `abs/path/exports` and was resolved as a workspace-relative join onto a nonsense location — the scan returned zero entries with no error.

This PR drops the leading-slash strip so `resolveExplorerRoot` (which already validates workspace containment via `path.relative` + `startsWith("..")`) accepts absolute inputs verbatim. The trailing-slash strip is also tightened so a bare `/` does not collapse to empty.

Adds three tests:
- `normalizeExplorerRootDir preserves absolute paths`
- `resolveExplorerRoot accepts an absolute path inside the workspace`
- `resolveExplorerRoot rejects an absolute path outside the workspace`

## Test Plan

- [x] `python -m unittest common.tests.test_assembly_composition -v` from `scripts/` — 9/9 pass
- [x] `node --test lib/cadDirectoryScanner.test.mjs` from `explorer/` — 10/10 pass
- [x] downstream end-to-end: regenerated a 10-instance per-leg assembly STEP and confirmed each child node in `topology.json`'s `assembly.root.children[].displayName` is now the per-instance name, and the dev server with absolute `EXPLORER_ROOT_DIR=/<workspace>/exports` lists the entries (previously zero).

### DIFF
--- a/.agents/skills/cad/explorer/lib/cadDirectoryScanner.mjs
+++ b/.agents/skills/cad/explorer/lib/cadDirectoryScanner.mjs
@@ -32,7 +32,7 @@ function encodeUrlPath(repoRelativePath) {
 
 export function normalizeExplorerRootDir(value = DEFAULT_EXPLORER_ROOT_DIR) {
   const rawValue = String(value ?? "").trim();
-  const slashNormalized = rawValue.replace(/\\/g, "/").replace(/^\/+/, "");
+  const slashNormalized = rawValue.replace(/\\/g, "/");
   const normalized = path.posix.normalize(slashNormalized);
   if (!normalized || normalized === ".") {
     return DEFAULT_EXPLORER_ROOT_DIR;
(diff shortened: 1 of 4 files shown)


earthtojake:
These look good to me, will clone and test locally then merge in

earthtojake:
Hi @arronzhang, I had a large change that i just pushed so included these fixes on that branch. Closing this branch (apologies for not merging it in!)

PR #12, opened 2026-05-21
feat(viewer): copy CAD view state

tolatolatop:
## Summary\n\nAdds a CAD Explorer toolbar action for copying the current review context as structured text. The clipboard payload includes a short human-readable summary followed by a JSON object with:\n\n- selected file / CAD path / render format\n- camera perspective\n- selected assembly parts and topology references\n- copyable @cad refs\n- hover IDs where available\n- hidden parts and expanded tree nodes\n- clip, theme, layout, and browser URL state\n\nThis is intended to let users share their exact CAD review context without relying on screenshots.\n\n## Testing\n\n- \n- \n- \n- Playwright smoke test against an assembly STEP confirmed the copied payload contains , camera perspective, selected part , and the corresponding  ref.\n


tolatolatop:
Closing at author's request; keeping the fork branch for local experimentation.

PR #13, opened 2026-05-22
WIP: feat(viewer): copy CAD review state URL

tolatolatop:
## Summary

Adds Copy/Paste View State actions to CAD Explorer so a user can share and restore a CAD review context without relying on screenshots.

The copied clipboard text contains a short human-readable summary followed by a versioned JSON payload. Paste restores the matching file when available, camera perspective, selected parts/references, hidden/expanded assembly state, clip settings, theme settings, and layout state.

This complements existing session persistence: session storage restores a local browser session, while view state is portable across chats, issues, machines, and reviewers.

## Why

When reviewing generated CAD with another person or an agent, screenshots are often not enough: camera angle, selected topology, hidden parts, clipping, and theme settings are lost. A portable view-state payload makes review context reproducible.

## Demo

Copy and paste CAD Explorer view state

## Scope

- Adds Copy View State and Paste View State toolbar actions.
- Adds a `CadExplorerViewState` clipboard document shape with schema/version constants.
- Keeps the payload plain text: summary first, JSON second.
- Does not add generated CAD artifacts, backend services, or external dependencies.

## Verification

- `node --test --experimental-default-type=module lib/workbench/viewState.test.js`
- `npm run test -- --runInBand`
- `npm run build`

### DIFF
--- a/viewer/docs/cad-review-state-url.md
+++ b/viewer/docs/cad-review-state-url.md
@@ -1,0 +1,103 @@
+# CAD Review State URL Payload
+
+This document defines the state that should be shared when a CAD Explorer user copies a review URL. The goal is to reproduce the reviewed model context without copying raw browser session storage.
+
+## Goal
+
+A copied review URL should allow another viewer session to open the same CAD file and restore the review-relevant 3D context: camera, selected entities, visibility, active assembly context, and clipping.
+
+The URL payload is not intended to mirror every transient UI state. Local persistence such as sessionStorage remains responsible for private, transparent restoration on the same browser.
+
+## Priority model
+
+- **P0 required**: missing this breaks review-state reproduction.
+- **P1 recommended**: improves semantic fidelity and diagnostics, but the core 3D review can still open without it.
+- **P2 optional**: useful for polish or format-specific workflows, but should not dominate the URL payload by default.
+- **Do not share**: transient or local-only state.
+
+## State inventory
+
+| Area | Parameters | Priority | Why it matters | URL placement |
+| --- | --- | --- | --- | --- |
+| File identity | `file.key`, `file.cadPath` | P0 | Selects the model the review state applies to. All camera, selection, and visibility IDs are meaningless on the wrong file. | Keep `file` readable in the URL; include both values in `view`. |
+| File metadata | `file.renderFormat` | P1 | Helps interpret format-specific state and validate restore behavior. | `view.file.renderFormat` |
+| File validation | content/topology/render signatures | P1 | Detects stale links or same-name/different-content models before applying IDs incorrectly. | Future `view.file.signatures` |
+| Camera | `position`, `target`, `up` | P0 | The minimum 3D viewpoint. | `view.camera.perspective` |
+| Camera metadata | `modelKey`, `sceneScaleMode`, `coordinateSystem` | P0 | The same numeric camera can render differently if model binding, scale mode, or coordinate system differ. | `view.camera.perspective` |
+| Projection | projection type, zoom/FOV | P1/future P0 | Needed once the viewer supports orthographic/perspective switching or explicit projection settings. | Future `view.camera.projection` |
+| Scene mode | `explorerMode` | P2/future | The 0.2 workspace does not expose a durable scene-mode state yet. Add this only when the viewer has a concrete mode to read and restore. | Future `view.scene.explorerMode` |
+| Scene mapping | `selectedRenderPartIdByAssemblyPartId` | P0 | Restores the render mesh that backs an assembly-part selection. | `view.scene.selectedRenderPartIdByAssemblyPartId` |
+| Scene path | active assembly path / expanded assembly context | P0/P1 | Required when assembly expansion also represents the active review context; otherwise it is sidebar context. | `view.assembly.expandedAssemblyPartIds` |
+| Selection | `selectedPartIds`, `selectedReferenceIds` | P0 | Defines what the review is pointing at. | `view.selection` |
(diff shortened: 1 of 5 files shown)


earthtojake:
This is very cool! Does it copy the entire session storage state for that file, including perspective?

tolatolatop:
Yes, it includes perspective (position/target/up) plus selected parts, references, hidden/expanded assembly state, clip settings, theme, and layout.

It's not a raw session storage dump though — view state is a curated portable snapshot designed for sharing across chats, issues, and machines, while session storage handles local browser restore. The clipboard payload is plain text (human-readable summary + versioned JSON) so it works anywhere you can paste text.

earthtojake:
how large do the states end up being? i wonder if we could just put them in query params and achieve the same effect, and potentially remove session storage altogether (excluding theme, which persists in local storage).

tolatolatop:
On the size: typical states are ~1.5 KB raw JSON / ~2.4 KB URL-encoded, heavy ones go past 12 KB. So query params with the full JSON payload are borderline.

But you're right that query params are the right approach — I was thinking about this at the wrong abstraction level. The receiver already has the file loaded, so we don't need to ship full part summaries and reference objects in the URL. All we need are pointers:

```
?file=model.step
&v=5.2,-3.1,8.8,0,0,0,0,0,1           ← camera (position x3, target x3, up x3)
&sr=o1.2.f1,o1.5.f3                     ← selected reference IDs
&sp=o1.2,o1.4                           ← selected part IDs
&hp=o1.10,o1.12                         ← hidden part IDs
&cp=z,0.5                               ← clip axis + position
```

Even with 10 hidden parts + 5 selected refs + clip + camera, the URL stays under 500 chars — smaller than the current clipboard payload and well within any URL limit.

The clipboard path is still useful for embedding view context inside issue comments or chat (human-readable summary + JSON), but URL should be the primary sharing path. It's a natural fit with the existing `?file=` and `?cadRef=` params.

On session storage: one thought — what if we keep it separate from view state? It handles transparent local restore with content-hash signatures (knows when generated CAD has changed and skips stale state), while view state is for explicit sharing between people. Different use cases, and the URL approach makes view state lighter anyway so there's less overlap. Curious what you think.

Happy to rework the PR around query params if you're on board with this direction.

tolatolatop:
Following up on the size point: putting the full state object directly into query params still feels a bit borderline, especially if we let it grow over time.

I reworked this around a smaller review-state payload instead. The URL carries the review context directly: the target file, camera/perspective, selected parts/references, hidden geometry, clipping, and the assembly/render context needed to make those selections line up again. From the reviewer’s perspective, the goal is that opening the link lands you on the same thing I was looking at, not just the same file.

I tried to keep the payload scoped to state that changes the reviewed view itself, and avoid transient UI state like hover, loading/error state, panel sizing, or theme preferences. My main concern is still that `view` can grow over time if we keep adding state to it, so I think we should treat it as a small review-state schema rather than a place to dump broader app state.

earthtojake:
yep i agree. i have a set of changes i'm about to push which refactors the viewer to be more modular and have clearer boundaries on what is state per file vs global. let me push that and then rebase this branch (i assume you're using codex/claude so should be able to handle gracefully).

two notes:
1. let's find a better place for those buttons since i'm not sure how frequently they'll be used
2. can we remove the "test" script from package.json and use the existing test framework?

tolatolatop:
Sounds good, I’ll wait for your viewer refactor and rebase after that.

Good luck with the changes!

earthtojake:
Hey, sorry for the delay here - i just pushed a big set of changes to the first stable release that is ready for contributors. If you still want to commit this PR, i'd suggest rebasing and asking codex/claude to migrate your changes over to the new repo structure! I appreciate you contributing so early, and apologies for not having the repo in a state that was ready for contributions.

tolatolatop:
Thanks, I rebased this onto the new 0.2 viewer structure and pushed the updated branch.

I also scoped the feature down to Copy view URL only: Paste view state was removed, and the package test script change is gone.

tolatolatop:
A bit more detail on what changed:

- Migrated the integration into the new `viewer/src/client/...` structure.
- Kept the URL-based sharing approach using a compact `cad-review-state` payload in the `view` query param.
- Preserved the current `dir` and relative `file` params so copied URLs reopen in the same workspace.
- URL restore now applies camera, selection, hidden parts, clip settings, render-part mapping, and active assembly inspection context.
- Invalid `view=` payloads now surface an error instead of silently falling back.
- Removed `explorerMode` from the P0 payload and marked it as future/P2 in the docs since the new viewer does not expose a durable mode state yet.

On button placement: I don’t have a strong settled idea yet. Since Paste is gone, the toolbar footprint is smaller, but `Copy view URL` may still be too low-frequency for the main floating toolbar. I’m happy to move it to a secondary or overflow location if there is a preferred place in the new viewer UI.

earthtojake:
@tolatolatop this is still pointing at main, can you point it at develop?

tolatolatop:
Done — I repointed this PR at `develop` and rebased/migrated the branch onto the latest `origin/develop`.

I also squashed the feature work down to a single commit on top of develop: `61a0ba2 feat(viewer): copy CAD review state URL`.

Validation on the rebased branch:

- `node --test src/client/workbench/viewState.test.js`
- `npm run build`

earthtojake:
Hey @tolatolatop unfortunately this branch has become pretty outdated, especially with the 0.4 release, and i'm not sure the feature makes sense anymore. Appreciate you being an early contributor, apologies the repo wasn't ready for contributions when you put this up! If you have reworks or other concepts in mind for this let me know. Closing for now.

PR #92, opened 2026-06-13
feat(viewer): feedback loop from CAD Viewer to agent

Celibistrial:
This was super useful to me so decided to make it a PR
## What

Adds a feedback loop from CAD Viewer back to the agent. In the Viewer, select a
face / edge / part, type what's wrong, and submit. Feedback persists as a sidecar
next to the model that the agent reads on demand and acts on — closing the
review loop without re-describing the part in chat.

## How it works

Submission writes, in the same `--dir` model root:

- `<model>.feedback.json` — array, one object appended per submission
- `<model>.feedback/<id>.png` — screenshot for that submission (path relative to the model root)

Item shape: `comment`, `references[]` (selected faces/edges/parts with
`copyText` selector tokens + `partId`), `camera`, `drawingStrokes`, `screenshot`.
The agent reads the JSON, opens the screenshot, acts on the comment + selectors,
then removes handled items (documented convention — no auto-cleanup).

## Changes

- **server** — `readFeedback` / `appendFeedback` on the local-fs backend:
  inside-root guard, atomic write, id-from-max, strict read that refuses to
  clobber a corrupt file; `GET`/`POST /__cad/feedback` (8 MB body cap for
  screenshots); skip `<model>.feedback/` dirs in the catalog scanner.
- **client** — `FeedbackPanel` (selection chips, comment, screenshot/annotation
  toggles, previous-feedback list) wired into `CadWorkspace` + a floating-toolbar
  button; captures selected references **and** parts, camera, and a screenshot.
  `captureScreenshotBlob()` added to the `CadViewer` imperative handle;
  `submitFeedback` / `fetchFeedback` store helpers.
- **docs** — `cad-viewer` SKILL "Review Feedback" section incl. the clear-on-done convention.
- gitignore the local feedback sidecars; backend + endpoint tests.

## Test plan

- `node --test viewer/src/server/{localAssetBackend,httpHandlers}.test.mjs` — 51 pass
- `npm --prefix viewer run build` — clean
- `scripts/bundle/bundle.sh --check` — up to date · symlink layout valid
- Manual: select parts → comment → submit → sidecar written with refs + highlighted screenshot; agent read + acted (blue / oval / fillet) end-to-end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,5 +35,9 @@ coverage
 *.stp.glb.lock
 tmp/
 
+# CAD Viewer feedback sidecars (local review artifacts)
+*.feedback.json
+*.feedback/
+
 # Local printer credentials
 /bambu-printers.json
--- a/skills/cad-viewer/SKILL.md
+++ b/skills/cad-viewer/SKILL.md
@@ -90,6 +90,47 @@ initializing. For a `"start"` result, probe `GET /__cad/server` on the base
 URL (e.g. `http://127.0.0.1:<port>/__cad/server`) until it returns HTTP 200
 before passing the `url` value to the Claude Preview tool.
 
+## Review Feedback
+
+Users can attach feedback to a model from inside CAD Viewer: select a face,
+edge, or part, open the feedback panel (message icon in the floating toolbar),
+type what is wrong, and submit. Each submission is persisted as a sidecar next to
+the reviewed model, in the same `--dir` model root:
+
+- `<model>.feedback.json` — JSON array, one object appended per submission.
+- `<model>.feedback/<id>.png` — screenshot for that submission (the `screenshot`
+  field is a path relative to the `--dir` model root, not inlined).
+
+Read `<model>.feedback.json` on demand (when the user says they left feedback, or
+when re-reviewing a model) and act on each item. Item shape:
+
+```json
+{
+  "id": "fb-0001",
+  "createdAt": "<ISO8601>",
+  "comment": "this fillet is too sharp",
+  "references": [
+    { "id": "f42", "label": "Face 42", "copyText": "f42",
+      "partId": "...", "entityType": "ADVANCED_FACE", "selectorType": "face" }
+  ],
+  "camera": { "position": {...}, "target": {...}, "up": {...}, "zoom": 1 },
+  "drawingStrokes": [ { "id": "...", "tool": "arrow", "points": [{"x":..,"y":..}] } ],
+  "screenshot": "<model>.feedback/fb-0001.png"
+}
+```
+
+Use `comment` for the user's intent and `references[].copyText` as the exact CAD
+selector token to locate the entity to change. Open the `screenshot` PNG for
+visual context, and use `camera` to reconstruct the exact view if needed. Empty
+`references` means the comment applies to the whole model.
+
+After you have acted on a feedback item, remove it from `<model>.feedback.json`
+(and delete its `screenshot` PNG) so it is not processed again on the next read.
+When every item is handled, delete the empty `<model>.feedback.json` and the
+`<model>.feedback/` directory. Submitting new feedback recreates them. There is
+no automatic notification or cleanup — feedback is only read when asked, and only
+cleared by you once resolved.
+
 ## References
 
 - Read `references/development.md` when the user asks to modify, debug, or
--- a/skills/cad-viewer/SKILL.md
+++ b/skills/cad-viewer/SKILL.md
@@ -100,6 +100,10 @@ the reviewed model, in the same `--dir` model root:
 - `<model>.feedback.json` — JSON array, one object appended per submission.
 - `<model>.feedback/<id>.png` — screenshot for that submission (the `screenshot`
   field is a path relative to the `--dir` model root, not inlined).
+- `<model>.feedback/<id>.strokes.json` — full annotation strokes for that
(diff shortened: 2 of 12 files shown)


earthtojake:
@Celibistrial this is great, thanks -- I actually added the "copy face/edge ID(s)" workflow to the app intentionally, but don't think it's obvious enough atm so this is a good improvement.

Only minor flag is that the drawing strokes can add a lot of data to context, and the strokes are already included in the screenshot.

Could you share some screenshots of what the workflow looks like for you on this branch?

Celibistrial:
Github has an unusually low limit so had to compress it a lot
https://github.com/user-attachments/assets/e908b70d-1faf-4754-b0fe-425f459a59d9
thats fair that the screenshot contains the sketch ,however for some complex drawings the screenshot MAY not be enough , made a change so that the agent optional reads the full strokes if it deems it necessary

earthtojake:
Oh interesting, how does Claude read the submitted feedback? Why do you prefer this flow vs just copying references back into the Claude CLI and adding context there?

Btw i started a group chat for users / contributors on x here if you want to join: https://x.com/i/chat/group_join/g2065804662922440796/S0qTlclCpv

Will setup a discord soon

Celibistrial:
> how does Claude read the submitted feedback?


The submission writes a sidecar next to the model in the --dir root — <model>.feedback.json (plus <id>.png screenshot and now <id>.strokes.json). The cad-viewer SKILL tells Claude to read that file when I say I left feedback or when re-reviewing.

Naming convention: same basename + extra extension, same directory.

eg:
ats_tec15_46_032.step              ← main artifact
ats_tec15_46_032.step.feedback.json   ← sidecar (feedback)
>  Why do you prefer this flow vs just copying references back into the Claude CLI and adding context there?


It's very convenient for multi part models , (i can stack up 5-6 feedbacks of different bodies / components and have claude go through them in one pass without any issues struggling with defining which feedback was for which part)

Copy-pastin references is not a good UX

earthtojake:
Yep makes sense, I have a few thoughts, where's the best place to discuss with you?

Main thing I'd like to avoid is writing feedback files to disk because cleanup etc can be messy

Celibistrial:
The feedback files are cleared once the agent deals with it. Github works best for me to discuss stuff

earthtojake:
@Celibistrial i did some research and codex + claude code have deep link features:

https://code.claude.com/docs/en/deep-links
https://developers.openai.com/codex/app/commands#deep-links

Is it worth investigating the UX for hooking into those deep links for the active thread?

Also, semi-related q, but how much value are you getting out of the file explorer part of the viewer? It adds a lot of bloat and blocks the viewer from being loaded as a static html file (has to be run as a server to read the user's file tree)

earthtojake:
https://github.com/earthtojake/text-to-cad/issues/97

Celibistrial:
I don't think they fit the in-session case cleanly 

Both schemes are built to cold start a session, not to push into a running one, at least from what i understand

earthtojake:
I did some more digging @Celibistrial and found these two options that should work for claude and codex:

Channels for Claude (can add an mcp server to the viewer server): https://code.claude.com/docs/en/channels
SDK for Codex (I believe Claude has an agents SDK too, not sure how similar they are): https://developers.openai.com/codex/sdk

Viewer links could include agent+thread IDs for comms back to the right thread.

This way the feedback loop would directly prompt the thread which would be great UX!

Not sure what the fallback would be for non-Codex/Claude agents.

Celibistrial:
Channels for claude does work , codexes SDK I don't think supports pushing directly into live session.

I don't think we should implement this change (as far as I understand this is meant to be agent agnostic)

Also sorry for the spam earlier , been testing an automated bot that has been misbehaving

Celibistrial:
Okay nvm , codex does indeed support it. Still no for gemini cli(its not like MCP, the session will have to be started from the app server)

Not sure if its worth it, what do you think?

earthtojake:
I think the better long-term solution here is to fork something like emdash.sh and integrate the CAD viewer into an electron app that actually wraps a bunch of app servers in a consistent interface!

That way the viewer is part of the thread interface so feedback loop is tighter. Wdyt?

earthtojake:
A good stepping stone might be an MCP server that handles push as well as pull, pretty sure this is supported in the MCP standard. Then the viewer's backend could stand up the MCP server for comms back to the thread!

earthtojake:
Closing this for staleness rather than for the idea, and thanks for building it.

The viewer-to-agent handoff shipped by a narrower route in the meantime: selecting geometry produces a copyable reference that carries its own file and occurrence path (`bracket#o1.2.f6`), so a ref pasted into a prompt tells the agent which model and which face without a panel in between. That covers a good part of what `FeedbackPanel.jsx` was reaching for, and the branch has drifted two months behind a viewer that changed underneath it.

If you still want the annotated-feedback part, which the copy flow does not do, a fresh PR against current `develop` would be easier than reviving this one.

earthtojake:
Reopened. I closed this by mistake a few minutes ago while tidying the PR list, and the closing note above was not a decision anyone made. Sorry for the noise. Nothing about the branch changed.

earthtojake:
@Celibistrial closing this PR as we are actively working on a desktop app that will more tightly integrate feedback from CAD visualization back to agents. Thanks for the work here!

PR #118, opened 2026-06-26
[codex] Add Burr skill

BrianLYS:
## Summary

Adds a self-contained `burr` skill that puts Burr design-rule checking inside a STEP-first parametric CAD generation and repair loop.

The workflow emits Burr design data alongside generated CAD, requires explicit rulepack selection, reads `burr.receipt.v2` outcomes, treats `incomplete` as a blocked trust claim, and uses structured repair guidance before rerunning affected source and geometry checks.

Burr lives at `skills/burr` with the other supported skills and is included in normal Skills CLI and provider-plugin discovery.

## Why

Visual review and geometry inspection do not prove that declared design intent was checked. Burr adds machine-readable design-rule receipts and repair guidance for declared metadata such as holes, bosses, insert pockets, slots, bearing seats, counterbores, spacing envelopes, and feature-pair rules.

Concrete example: [a T-slot slider with 0.10 mm side clearance looks valid in CAD but may bind](https://github.com/earthtojake/text-to-cad/pull/118#issuecomment-5315184735). Burr rejects it against the selected 0.20–0.35 mm range; after changing the source to 0.25 mm and regenerating the STEP, all 14 checks pass.

Burr does not claim structural safety, process capability, tolerance compliance, material suitability, or print success, and the skill does not depend on sibling skills at runtime.

## Packaging

- Listed with the supported skills in README and docs
- Discovered as `burr` by the shared installer and root plugin package
- Self-contained under `skills/burr`

## Validation

Passed:

- `scripts/test/test.sh` (full JavaScript and Python suite)
- `scripts/dev/setup-symlinks.sh --check`
- `scripts/test/test-global.sh` (85 tests)
- `scripts/test/test-docs.sh`
- `scripts/bundle/bundle.sh --check`
- `scripts/release/check-version.sh`
- `git diff --check upstream/develop...HEAD`
- supported discovery check confirming `scripts/utils/list-skills.sh` includes `burr`


BrianLYS:
Hello @earthtojake , this adds https://github.com/fraylabs/burr skill to text-to-cad.
Burr is a linter with design check rules for building mechanically sound parts.
Currently it supports build123d.

earthtojake:
@BrianLYS this looks great! Do you think this should be added to the core $cad skill, or is it too specific of a workflow?

If you want to keep discussing jump in Discord: https://discord.gg/nxqvbJ73P

BrianLYS:
Thank you for reviewing @earthtojake !

I think `cad-burr` should stay separate for now.

Burr is still an opinionated workflow: metadata, rulepacks, receipts, and repair loops. I don't think every `$cad` task  should pay that complexity upfront yet.

We already have real examples where Burr helps, and I’m working on a small gallery to showcase them better. Once the examples feel broad enough beyond holes/slots, we can decide what should fold into core `$cad`.

BrianLYS:
Ok hopping on discord! See ya there 🫡

BrianLYS:
burr-clearance-comment


Here’s a concrete example of what Burr catches:

This T-slot slider looks fine in the CAD render, but its 0.10 mm side clearance is below the selected 0.20–0.35 mm range and may bind.

A binding fit is too tight and may stick or jam once friction and print variation are involved. A sliding fit keeps controlled clearance so the carriage can move while the rail still guides it.

After changing the source to 0.25 mm clearance and regenerating the STEP, Burr passed all 14 checks.

PR #120, opened 2026-06-27
Release 0.4.0

earthtojake:
## Release 0.4.0

Accumulated CAD-skill, viewer, and Python-core release: **372 commits over `develop`** across 30 merged PRs, `1571 files changed, +156076 / -62891`. `develop` (through release 0.3.10) is merged in, so this branch is a strict superset — there are zero commits on `develop` that are not here. The most recent sync resolved two conflicts from [#192](https://github.com/earthtojake/text-to-cad/pull/192): it fixed `viewer/src/server/httpHandlers.mjs`, which this branch deleted along with the rest of the Node backend. The underlying crash (`decodeURIComponent` throwing on `/__cad/foo%zz.step`) has no analogue in the Python replacement — `urllib.parse.unquote` does not raise on a malformed escape — and that is pinned by `viewer/server_py/tests/test_malformed_percent_encoding.py` rather than left as an assumption.

`plugins/cad/VERSION` is `0.4.0` and derived metadata is stamped from it. Do not hand-edit the duplicates; the `Release` workflow owns the real version bump.

---

### Python core

| PR | Change |
| --- | --- |
| [#128](https://github.com/earthtojake/text-to-cad/pull/128) | Publish **cadgen** (formerly `cadpy`) to PyPI with a public/internal API split |
| [#133](https://github.com/earthtojake/text-to-cad/pull/133) | Split the STEP CLI into `scripts/gen` (build render packages) and `scripts/export` (write files); add `scripts/artifact` debug CLI |
| [#140](https://github.com/earthtojake/text-to-cad/pull/140) | cadgen edit-path perf: AST closure hash, classification cache, single BREP serialize |

### Viewer

| PR | Change |
| --- | --- |
| [#123](https://github.com/earthtojake/text-to-cad/pull/123) | **Python viewer backend** (`viewer/server_py`) replaces the Node server |
| [#135](https://github.com/earthtojake/text-to-cad/pull/135) | Simplify launch to a single `start` command on port **3245** (0xCAD) |
| [#150](https://github.com/earthtojake/text-to-cad/pull/150) | Require a valid CAD Python backend at startup |
| [#136](https://github.com/earthtojake/text-to-cad/pull/136) | Exploded view redesign: step-based scrub slider, auto-explode, authoring, trails |
| [#138](https://github.com/earthtojake/text-to-cad/pull/138) | Renderer consolidation: shared camera kit, cadjs/implicitjs dedup, implicit ref-parity |
| [#137](https://github.com/earthtojake/text-to-cad/pull/137) → [#144](https://github.com/earthtojake/text-to-cad/pull/144) | Package rendering: cid-keyed `InstancedMesh` landed first, then **replaced** by a single shared-geometry path with instancing removed. #144 is the shipped design — one render path, per-occurrence `Mesh` over shared component geometry |
| [#141](https://github.com/earthtojake/text-to-cad/pull/141) | Instanced interaction: zoom-to-fit selection + incremental hover |
| [#147](https://github.com/earthtojake/text-to-cad/pull/147) | Consistent pan/zoom at every zoom level (anchored zoom-to-cursor pivot) |

### Snapshot

| PR | Change |
| --- | --- |
| [#143](https://github.com/earthtojake/text-to-cad/pull/143) | Snapshot unification + implicit-model snapshots through the unified pipeline |

### Generation reliability, snapshot unification, CLI output

Three merges after #204's cut, each measured rather than asserted.

| Area | Change |
| --- | --- |
| [#204](https://github.com/earthtojake/text-to-cad/pull/204) | **Contended builds report the peer instead of stalling behind it.** `deadline_ms`/`Contended` existed, were tested and had no production caller, so every acquire blocked forever silently — `cad gen --force` behind a peer's lock sat 35s emitting zero bytes. One contended model also froze the viewer's builds for *every* model (an unrelated, already-current model measured **31.99s → 0.0s**), because a POST runs in the one serial warm worker. An export additionally clobbered a live build's progress record and erased the learned phase weighting; the generator run writes its own record now |
| direct merge | **Snapshot CLIs unified.** One implementation (`cadgen.snapshot_cli`), six front doors. CAD's entrypoint **1224 → 53 lines**; DXF's hand-written shell and implicit's standalone **978-line Node CLI** both deleted. Formats gate per skill (cad: `.step.py`/`.step`/`.stp`/`.3mf`/`.glb`/`.stl`; implicit-cad; dxf; and **urdf/srdf/sdf, which gain snapshot for the first time**). `--appearance` → `--theme`, hard, 420 occurrences. A new render-only `snapshot` theme (Workbench Light minus grid, origin axis and shadows) is the default and is absent from the viewer's picker |
| direct merge | **CLI output tidied for agents.** The parts inventory (`snapshot --mode list`) was the one O(n) payload: **294 KB / 12,619 lines → 94 KB / 1 line, 3.1×, ~50k tokens** — `id`/`occurrenceId`/`ref` were one value three times (600/600 parts), `label` duplicated `name` (600/600), coordinates carried 16 significant figures for millimetres. stdout is now the result on every tool (`gen` printed nothing there at all), four previously-silent CLIs narrate, and errors are compact instead of ~60-line tracebacks |

**Two behaviour changes worth knowing before merge:**

- **Non-STEP snapshots now render orthographic, not perspective.** `renderJobContext` forced every non-STEP source to perspective regardless of theme, and the tight-frame pass that fits the canvas to the geometry is orthographic-only — so it silently never ran and every mesh, drawing, implicit and robot snapshot sat in empty space. Projection is a theme trait for all formats now, matching the viewer's own U1 rule. Any baseline images predating this will differ.
- **The `--mode list` payload dropped `id`, `occurrenceId` and `label`.** `ref` is the survivor; it is the form that feeds `--focus`/`--hide`/`inspect`.

Design docs: `design/snapshot-cli-unification.md`, `design/cli-output-tidy.md`. New policy tests pin the theme/display parity with the viewer, the CLI stream contract, and the payload budget.

### Skills

| PR | Change |
| --- | --- |
| [#127](https://github.com/earthtojake/text-to-cad/pull/127) | Rearchitect URDF/SRDF/SDF skills to direct XML authoring with bolstered validators |
| [#130](https://github.com/earthtojake/text-to-cad/pull/130) | DXF pipeline parity with STEP: `.dxf.py` generators, drawing packages, viewer auto-regen |

### Backport

| PR | Change |
| --- | --- |
| [#179](https://github.com/earthtojake/text-to-cad/pull/179) | Backport the release 0.3.10 fixes (#158, #169, #171, #172, #174, #176) |

---

## Structural changes reviewers should know about

These are renames and deletions rather than edits, so the diff overstates the churn:

- **`packages/cadpy` + `packages/cadpy_metadata` → `packages/cadgen`.** The old paths are gone from this branch; `develop` still has them.
- **Entry generators renamed `<name>.py` → `<name>.step.py`** — 172 pure renames under `models/`. The viewer scans for `.step.py`, not every `.py`.
- **Parameter sidecars renamed `.step.js` → `.params.js`.** Note this diverges from `develop`, which kept `.step.js`; see the known gap on #177 below. (The 43 `.implicit.js` models are identical on both branches.)
- **Committed monolithic artifacts removed** — 162 `.glb` and 139 `.step` files. `gen_step` no longer writes STEP; artifacts are generated on demand into content-addressed component-GLB packages, and `--step` exports a file when you actually want one.
- **`viewer/src/server/*.mjs` deleted** in favour of `viewer/server_py/`.
- **`scripts/bundle/skills/bundle-{sdf,srdf,urdf}.sh` folded** into `scripts/bundle/lib/vendor.sh`.

---

## Known gaps

Two release 0.3.10 PRs were deliberately **not** backported, so they are still absent here:

- **[#175](https://github.com/earthtojake/text-to-cad/pull/175)** (viewer startup docs). This branch already documents its real launcher (`start`, port 3245) rather than develop's `serve`/4178, so the docs half was moot. What is still missing is the functional half: `viewer/server_py/server.py` binds with a bare `ThreadingHTTPServer((host, port))` and has no scan-forward, so a busy port fails with `EADDRINUSE` instead of picking the next one. Upstream's `listenWithPortFallback` is the Node equivalent of the fix.
- **[#177](https://github.com/earthtojake/text-to-cad/pull/177)** (`models/` file-type policy). Its `test_models_directory_policy.py` rejects this branch's 16 `.params.js` sidecars, so neither the policy docs nor the enforcing test were taken. Resolving this needs the sidecar naming to converge with `develop` (or the allow-list widened). The `.gitignore` rules keeping review media out of `models/` **were** taken — they stand alone and this branch tracks no media there.

Also still open against this branch: **[#142](https://github.com/earthtojake/text-to-cad/pull/142)** (STEP "Open in \<CAD app\>" context-menu integrations).

---

## Verification

CI is green on the current tip (`843a9884`): **Test** and **Version Check** both pass on [run 30492659502](https://github.com/earthtojake/text-to-cad/actions/runs/30492659502).

Locally against the same tip: cadjs 421, implicitjs 56, viewer 246, `scripts/test/test-python.sh` 557 across all skill suites, `scripts/test/test-global.sh` 12, `scripts/test/test-docs.sh`, `npm --prefix viewer run build`, `scripts/bundle/bundle.sh --check`, and `scripts/dev/setup-symlinks.sh --check` all pass.


pranavanand:
LGTM

PR #122, opened 2026-06-28
Preserve GLB material opacity in CAD viewer

Marcus-QL-Zhu:
## Summary

- Preserve GLB material opacity while converting GLB meshes into cadjs meshData parts.
- Apply source part opacity when building CAD scene materials, so STEP/GLB materials with alpha render translucent in the browser viewer.
- Auto-render parts individually when source material opacity differs, even when `renderPartsIndividually` is false, to avoid silently flattening away part-level opacity.

## Why

Some STEP-to-GLB pipelines preserve transparency as `baseColorFactor[3]` / loaded Three.js material opacity, but do not always set glTF `alphaMode`. FreeCAD can display the STEP material transparency correctly, while CAD Explorer could lose opacity after converting GLB nodes into internal meshData because only RGB/hex color was carried forward.

## Quick Reproduction

1. Generate or open a STEP-derived GLB whose material has `pbrMetallicRoughness.baseColorFactor[3] < 1`, for example `0.2`, but no explicit `alphaMode: BLEND`.
2. Load the same source model in a CAD tool such as FreeCAD and confirm the STEP material appears translucent.
3. Load the STEP/GLB through CAD Explorer.
4. Before this change, the browser viewer renders the part as opaque because the GLB-to-meshData conversion keeps RGB/hex color but drops material opacity.
5. After this change, the part-level opacity is preserved and applied to the generated Three.js material.

A focused unit-level reproduction is included in the tests by using GLB material metadata with alpha but without `alphaMode`, then verifying that `meshData.parts[*].opacity` and the final CAD scene material opacity are preserved.

## Tests

```bash
node --test packages/cadjs/src/common/cadScene.test.js packages/cadjs/src/lib/render/glbMeshData.test.js
```

Result: 27 tests passed.

### DIFF
--- a/packages/cadjs/src/common/cadScene.js
+++ b/packages/cadjs/src/common/cadScene.js
@@ -599,6 +599,11 @@ function sourceColorForPart(THREE, part, meshData) {
   return readSourceColor(THREE, part?.color || meshData?.sourceColor);
 }
 
+function sourceOpacityForPart(part, fallback = 1) {
+  const opacity = Number(part?.opacity);
+  return Number.isFinite(opacity) ? clamp(opacity, 0, 1) : fallback;
+}
+
 function meshUsesPartSourceColors(meshData, parts) {
   const renderableParts = Array.isArray(parts) ? parts : [];
   const partColors = renderableParts
@@ -608,6 +613,14 @@ function meshUsesPartSourceColors(meshData, parts) {
     return false;
   }
   return partColors.length !== renderableParts.length || new Set(partColors).size > 1;
+}
+
+function meshUsesPartSourceOpacity(parts) {
+  const renderableParts = Array.isArray(parts) ? parts : [];
+  return renderableParts.some((part) => {
+    const opacity = Number(part?.opacity);
+    return Number.isFinite(opacity) && clamp(opacity, 0, 1) < 0.999;
+  });
 }
 
 function emptyLineGeometry(THREE) {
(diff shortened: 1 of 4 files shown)


earthtojake:
Thanks for this @Marcus-QL-Zhu ! Mind sharing a screenshot with a before/after example? This is for GLB exports from STEP files, not the GLB artifacts used to render STEP files (behind the scenes, not exposed to users), right?

earthtojake:
Looks good, updated target to develop and merging

Marcus-QL-Zhu:
> Looks good, updated target to develop and merging

@earthtojake Hi, happy to contribute! Though this is my first time pushing PR for other's repo. What do I need to do now? Do you still need my screenshot? I can provide that after like 16 hours

earthtojake:
Can you rebase onto the develop branch? I checked this locally and it looks good, but the branch itself has a bunch of unnecessary changes included

Marcus-QL-Zhu:
OK I'll try

Marcus-QL-Zhu:
Rebased onto develop and force-pushed a clean branch. The PR now only includes the CAD viewer opacity fix and related tests. Please kindly check and let me know if any other modification needed! @earthtojake

earthtojake (approved):

earthtojake:
Merged, thanks @Marcus-QL-Zhu ! For some reason the repo didn't add you as a contributor, but will look into it and make sure your next contribution gets you added to the repo's contributors list.

PR #145, opened 2026-07-09
Fix formatting in SKILL.md for clarity

afianurul13:


earthtojake:
Hey @afianurul13 , where is this specific formatting recommended?

earthtojake:
Thanks for the contribution, @afianurul13 — I'm going to close this one, but let me explain why so it's useful for next time.

To answer my question above: nothing in this repo recommends the spaced-hyphen style. There's no markdownlint/prettier/vale config, and no dash rule in `CONTRIBUTING.md` or `AGENTS.md`. The em dashes here are intentional and used correctly as parenthetical dashes.

Two things would block it regardless of style preference:

- **It targets `main`.** `main` is publish-only in this repo — it's generated by the `Release` workflow, and PRs should target `develop`. Merging a commit directly into `main` breaks the next release: with no `Source commit:` trailer on the merge, `scripts/release/check-publish-source.sh` falls back to the merge commit's first parent, which isn't an ancestor of `develop`, so the publish gate fails. It's also why no CI ran here — `test.yml` only triggers on `develop`.
- **It would desync the published plugin.** On `main`, `plugins/cad/skills/cad/SKILL.md` is a materialized copy of the skill source (it's a symlink on `develop`). Changing only `skills/cad/SKILL.md` leaves the two 11 lines apart.

On consistency: `skills/cad/SKILL.md` isn't the only file using em dashes — 9 more remain across `skills/cad/references/`, `skills/cad-viewer/SKILL.md`, and `skills/dxf/SKILL.md`. So this would make the CAD skill inconsistent with its own reference set rather than more consistent.

Contributions are genuinely welcome — just branch from `develop` and target `develop`. Thanks again for taking a look at the repo!

PR #148, opened 2026-07-10
Update python version from 3.11 to 3.12.

nathan-t4:
# Problem

Installing `requirements-dev.txt` using a python3.11 venv causes an error since `viewer/moveit2-server` requires python3.12

https://github.com/earthtojake/text-to-cad/blob/622c258fbdea64614056bfe33f3d3f183021a374/viewer/moveit2_server/pyproject.toml#L9

In fact, the github action already uses python 3.12.

https://github.com/earthtojake/text-to-cad/blob/622c258fbdea64614056bfe33f3d3f183021a374/.github/actions/setup-deps/action.yml#L12

## Steps to reproduce error

Following the CONTRIBUTING.md guidelines
```
uv venv --python 3.11
source .venv/bin/activate
uv pip install -r requirements-dev.txt
```
produces the error:
```
  × No solution found when resolving dependencies:
  ╰─▶ Because the current Python version (3.11.14) does not satisfy Python>=3.12 and moveit2-server==0.3.8
      depends on Python>=3.12, we can conclude that moveit2-server==0.3.8 cannot be used.
      And because only moveit2-server==0.3.8 is available and you require moveit2-server, we can conclude that
      your requirements are unsatisfiable.
```

# Solution

Update the contributing docs and every reference from python 3.11 to 3.12.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ git switch -c my-change
 Create the repo-local Python development environment:
 
 ```bash
-python3.11 -m venv .venv
+python3.12 -m venv .venv
 ./.venv/bin/python -m pip install --upgrade pip
 ./.venv/bin/python -m pip install -r requirements-dev.txt
 ```
--- a/packages/cadpy/pyproject.toml
+++ b/packages/cadpy/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "cadpy"
 version = "0.3.8"
 description = "Shared STEP/GLB topology artifact generation runtime for CAD skills."
-requires-python = ">=3.11"
+requires-python = ">=3.12"
 dependencies = [
   "build123d",
   "cadquery-ocp",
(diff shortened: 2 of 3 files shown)


earthtojake:
Reviewed — LGTM. ✅

**Why this is correct:** `requirements-dev.txt` installs `viewer/moveit2_server` (already `requires-python >= 3.12`) alongside `cadpy`/`cadpy_metadata`, so a 3.11 venv can't resolve the dev environment at all, and CI (`.github/actions/setup-deps`) already runs Python 3.12. Aligning `packages/cadpy`, `packages/cadpy_metadata`, and the CONTRIBUTING venv instructions to 3.12 is the right fix, and touching only the source `pyproject.toml` files is sufficient on `develop` — the vendored copies under `skills/*/scripts/packages/` and `plugins/cad/` are symlinks to these sources. No release-stamped `version` fields are affected.

**Process notes:**
- Retargeted the base branch from `main` → `develop`: `main` is the publish-only release branch in this repo, and on `main` the vendored package copies are materialized files that would have gone stale. The diff is identical against both branches.
- The required `Test` / `Version Check` checks don't auto-run after a retarget, so they were run against this PR's exact head commit (374640b) via workflow dispatch: both passed — https://github.com/earthtojake/text-to-cad/actions/runs/29118527629

**Nit (follow-up, non-blocking):** the README badge still advertises `Python 3.11+`.

Thanks for the fix, @nathan-t4!

PR #154, opened 2026-07-15
docs(models): add verified codex demo CAD assets

Bukuto959:
- add reproducible Python sources for two demo models;
- add STEP/DXF/GLB outputs and topology metadata;
- add verified isometric/top-view snapshots for viewer and documentation use.

Testing:
- source/destination SHA verified for all 12 paths;
- development symlink layout check previously passed at the same unchanged HEAD;
- Git LFS integrity check passed;
- clean worktree and staged=0.


earthtojake:
Apologies for the slow review here — thanks for the patience.

I pulled this down and regenerated both models. The **CAD sources are solid**: both generators run cleanly under build123d 0.10.0 / ezdxf 1.4.3, and the regenerated STEP and DXF match the committed artifacts exactly — identical face/edge counts, bounds, and per-entity DXF geometry. Clean, readable parametric source.

Three things I'd like fixed before merge:

**1. `xiaoluo_display_base` accessory holes cut into the pedestal.**

`Hole()` with no `depth` drills through the entire part, not just the base plate. The centers sit 22.804 mm from the pedestal axis, so with r=2.5 the inner edge lands at 20.304 mm — inside the 22 mm pedestal radius, gouging two 1.7 mm channels down the full 18 mm pedestal wall. It's visible in the iso snapshot in this PR (the vertical scallops on the pedestal) and in the top view, where the pedestal outline runs straight through both lower holes. Fix with `depth=BASE_THICKNESS`, or move `ACCESSORY_Y` out past ~24.5 mm from the axis.

Separately, the D12 "socket" also runs all the way through the base rather than being blind — intended?

**2. The PNGs and topology JSON conflict with the policy in `models/README.md`.**

The cleanup policy there says not to commit `*.png` / `*.json` supplementary media or sidecar metadata, and not to commit "one-off timestamped review snapshots." These four files are both. They also aren't LFS-tracked — `.gitattributes` has no PNG rule, so `git check-attr` reports `filter: unspecified` and ~361 KB of raw binary would land in normal git history permanently. There are currently no PNGs anywhere under `models/`. Could you drop the three snapshots and `four_hole_plate_topology.json`? The topology is regenerable from `inspect refs` on demand.

**3. `four_hole_plate_topology.json` starts with a UTF-8 BOM**, so it fails both `json.load(encoding='utf-8')` and `JSON.parse`. Moot if it's dropped per #2, but it means the file was never read back after being written.

Minor:

- `models/README.md`'s Directory Map links a README for every sibling; `codex-demo/` would be the only directory under `models/` without one.
- No CI has run on this — it's a cross-repo PR, so the workflow is sitting behind first-time-contributor approval. Worth getting a green run before merge, since the Testing notes in the description are all self-reported and don't exercise the models.
- `four_hole_plate` is close to what already lives in `models/simple/` — would that be a better home than a new top-level directory?

earthtojake:
Closing this as out of date rather than unwanted, and sorry it sat so long.

The repo's model layout has moved on since July in three ways that this PR predates:

- Entries are named `<name>.step.py`. The Viewer scans for that suffix, so `four_hole_plate.py` would not be discovered.
- Generated artifacts live under a `__cadgen__/` package beside the source and are rebuilt on demand. The dot-prefixed `.four_hole_plate.step.glb` sidecars and `four_hole_plate_topology.json` are formats the current pipeline no longer reads; nothing on `develop` uses either.
- Exported `.step`, `.dxf` and snapshot PNGs are regenerated rather than committed.

The models themselves would still be welcome. Rewritten as `four_hole_plate.step.py` and `xiaoluo_display_base.step.py` under `models/step/parts/`, with the generated files left out, this would go in cleanly. Happy to look at that if you would like to open it.

PR #158, opened 2026-07-20
docs(bambu-labs): document TLS verification bypass mitigation

xiaolai:
> **Automated**: drive-by fix from [NLPM](https://github.com/xiaolai/nlpm), an NL artifact linter. Reviewed and reproduced before submission.

**Bug**: TLS certificate verification is disabled (`CERT_NONE`) at 4 call sites in `bambu_lan_print.py` with no in-place documentation of why this is safe.

**Evidence**: Confirmed all 4 `verify_mode = ssl.CERT_NONE` sites (lines 379-380, 943-944, 1152-1153, 1210-1211) are preceded by a call to `validate_local_host()`, which resolves the target and refuses non-private/non-loopback/non-link-local addresses unless `--allow-nonprivate-host` is passed.

**Fix**: Adds a short comment at each site cross-referencing `validate_local_host()` so the LAN-only mitigation for Bambu's self-signed printer certificates is documented in place, not only in the caller.

### DIFF
--- a/skills/bambu-labs/scripts/bambu_lan_print.py
+++ b/skills/bambu-labs/scripts/bambu_lan_print.py
@@ -376,6 +376,9 @@ def discover_printer_serial(
     validate_local_host(host, allow_nonprivate_host)
     context = ssl.create_default_context()
     if not tls_verify:
+        # Bambu printer firmware serves self-signed LAN certificates; validate_local_host()
+        # above restricts this to private/loopback/link-local hosts, so disabling verification
+        # here only trusts the printer's subnet, not its actual certificate.
         context.check_hostname = False
         context.verify_mode = ssl.CERT_NONE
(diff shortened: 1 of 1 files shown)


earthtojake (approved):
Verified all four `CERT_NONE` sites are guarded by `validate_local_host()`. Ran the bambu-labs suite (37/37) and the repo policy tests (7/7) against this head; comment-only change, no behavior impact, and `plugins/cad/skills/bambu-labs` is a symlink so no bundle refresh is needed.

Non-blocking follow-up: the comment states the private/loopback/link-local restriction unconditionally, but `validate_local_host()` returns early when `--allow-nonprivate-host` is set, so unverified TLS can still reach a public host in that case (the PR description states this caveat correctly). Worth a one-line amendment in a later pass.

PR #163, opened 2026-07-23
test(implicitjs): add missing unit tests for getViewerThemeValue

escmontiel03-boop:
Adds unit test coverage for `getViewerThemeValue` in `packages/implicitjs/src/lib/viewer/stageTheme.js`, which previously had no tests. Covers: preferring the explicit viewer theme value, falling back to `BASE_VIEWER_THEME`, and falling back to the caller-provided default for unknown keys.

Verified with `node --test src/lib/viewer/stageTheme.test.js` (3 passing).


earthtojake:
Thanks for this — the placement and wiring are right. Colocating under `src/` matches the 12 sibling `*.test.js` files, and `scripts/run-tests.mjs` picks the file up with no extra wiring. I ran the full package suite locally: 59/59 pass, including these three. CI is green now too.

One substantive thing before this goes in: **the tests don't currently guard the behavior that matters.**

I mutation-tested it by changing the `??` chain in `getViewerThemeValue` to `||`:

```js
return value || BASE_VIEWER_THEME[key] || fallback;
```

All three tests still pass. Every input in the PR is a truthy string, so the nullish-vs-falsy distinction — really the only interesting thing this three-line function does — goes unasserted.

That distinction is load-bearing here, because `BASE_VIEWER_THEME` is half numbers (`surfaceRoughness: 0.92`, `surfaceMetalness: 0.03`):

| input | `??` (current, correct) | `\|\|` (bug) |
|---|---|---|
| `{ surfaceMetalness: 0 }` | `0` | `0.03` |
| `{ surfaceRoughness: 0 }` | `0` | `0.92` |
| `{ surface: "" }` | `""` | `"#f4f4f5"` |

So a caller asking for a fully non-metallic or perfectly smooth surface would silently get the base value back instead. Could you add a case along these lines?

```js
test("getViewerThemeValue preserves falsy-but-defined viewer theme values", () => {
  assert.equal(getViewerThemeValue({ surfaceMetalness: 0 }, "surfaceMetalness", 1), 0);
  assert.equal(getViewerThemeValue({ surfaceRoughness: 0 }, "surfaceRoughness", 1), 0);
  assert.equal(getViewerThemeValue({ surface: "" }, "surface", "#000000"), "");
});
```

I verified that version locally: 60/60 with it in place, and it correctly fails (`not ok 4`) against the `||` mutation, so it has real teeth.

Minor, non-blocking: three separate `test()` blocks for one three-line function is finer-grained than the surrounding style — `packages/cadjs/src/lib/viewer/stageTheme.test.js` groups related helper assertions into a single named block. Fine to leave as-is if you prefer.

Once the falsy case is added, this is good to merge. The branch will also need an update from `develop` before it can go in.

warun7:
The requested falsy-value test has been added on top of the PR branch. The contributor's branch was already updated from develop (merge `73496f38`), so that requirement is covered.

Since we do not have write access to the contributor's fork, the ready changes live at `warun7/text-to-cad:test/stage-theme-viewer-value` (commit `9b65044c`, one commit on top of the contributor's tip). With maintainer edits enabled you can fetch and update the PR branch directly, or pass the commit hash to the contributor.

What was added, exactly as suggested:

```js
test("getViewerThemeValue preserves falsy-but-defined viewer theme values", () => {
  assert.equal(getViewerThemeValue({ surfaceMetalness: 0 }, "surfaceMetalness", 1), 0);
  assert.equal(getViewerThemeValue({ surfaceRoughness: 0 }, "surfaceRoughness", 1), 0);
  assert.equal(getViewerThemeValue({ surface: "" }, "surface", "#000000"), "");
});
```

Verified locally:
- Full implicitjs suite: 351 tests, 350 pass, 1 skipped, 0 fail. The new test passes.
- Mutation check: with `??` changed to `||` in `getViewerThemeValue`, the new test fails, so it has real teeth.

@earthtojake the branch is ready either way you want to land it.

earthtojake:
@escmontiel03-boop the falsy-value test now exists as warun7/text-to-cad@9b65044c (thanks @warun7), one commit on top of your current tip, but it is not on this PR branch yet. Cherry-pick that commit or copy the test block from my review and push. Once it lands here and CI is green I'll merge.

earthtojake:
Closing. packages/cadgen-js/src/lib/viewer/stageTheme.test.js on main now covers getViewerThemeValue, and the falsy-value case never landed here. Thanks for starting it.

PR #169, opened 2026-07-28
Viewer: reduce load under software WebGL

n8guru:
## Summary

- detect common software WebGL renderers through the renderer context
- cap software-rendered sessions at 1x pixel ratio
- disable shadow maps and key-light shadows only for software rendering
- preserve existing hardware-rendered behavior

## Validation

- packages/cadjs: 383 tests passed
- viewer: 386 tests passed, 4 skipped
- production Viewer build passed under Node 22
- development symlink layout check passed

Independent Forage review task 7836 was dispatched but had not completed within the bounded wait when this PR was opened.


earthtojake:
Thanks — this looks good. Verified locally: SwiftShader/Metal detection behaves correctly, hardware rendering is untouched, and a sustained orbit under software WebGL goes from ~1.2 fps to ~5.5 fps at DPR 2. Merging once CI is green.

PR #180, opened 2026-07-30
snapshot: reject render-asset cache hits across job sources

rootsbymenda:
## Summary

Follow-up 2 from #172 — the runtime assertion: a cache hit whose recorded source differs from the
requesting job's now raises instead of silently serving another source's bytes.

`#172` made a collision *unlikely* by keying URLs on file identity; nothing yet *detected* one. The
batch renderer keeps one Chromium page across jobs, so every cadjs cache is page-lifetime — any
future path that maps two sources onto one URL would still silently reuse pixels. This closes that
class: every cache entry is tagged with the source it was populated under, and a cross-source hit
throws with both sources named.

## Mechanism

- New `packages/cadjs/src/lib/renderAssetSourceScope.js` holds a module-scoped source identity and
  the assertion. `loadSource` (`common/source.js`) declares the scope from `job.resolved.inputPath`
  before its loads and restores the previous scope in a `finally` — one job can never leave its
  identity attached to another caller's loads.
- The assertion is installed in **`stepRenderAssetClient.js`** — the client the snapshot bundle
  actually ships — in `loadCached` and the array-buffer path, and in `renderAssetClient.js` for
  symmetry. Claims are **recorded on populate** and **released when a failed or aborted load deletes
  its cache entry**, so an error never poisons a URL for a later legitimate source.
- `peek*` readers return a miss (`null`) on a cross-scope entry rather than throwing — peeks are
  advisory; loads are where a wrong reuse would become a wrong image.
- `loadSelectorRuntime` / `loadDisplayEdgeRuntime` swallow missing-sidecar errors as before, but now
  **rethrow scope violations** — those two catch blocks would otherwise have silently discarded the
  assertion on the exact channels the stale-pixel bug lived in.

## Why `resolved.inputPath` is the identity

It is the same identity the server routes by, one step finer: `__main__.py` derives
`root_path = input_path.parent.resolve()` (`:870`, `:875`), so rootPath is a pure function of
inputPath, and inputPath additionally distinguishes two inputs in one directory forced onto one URL
— which is precisely the "future re-collision" this assertion exists to catch. Same-stem pairs
(`gear.py` / `gear.step`) resolve to one logical input, so their legitimate shared fetch is
preserved.

## One documented shape is affected — flagging it rather than hiding it

`packages/cadjs/docs/render-pipeline.md` listed `resolved.glbUrl` and `resolved.inputPath` as
independent alternatives. Under this change, a `resolved` packet carrying a source URL without
`inputPath` is rejected (fail closed) — a page-lifetime cache entry under an unidentified source is
the hole this PR exists to close. No in-repo caller uses that shape (the docs hero renderer passes
no `resolved` packet at all and is unaffected — verified at runtime against the real hero STEP GLB).
I updated the doc to state the contract; if you'd rather the guard tolerate that shape instead, say
so and I'll adjust.

## Tests

- `stepRenderAssetClient.test.js` is new — that module previously had no test file.
- **Composition-level tests drive the real wiring** (`loadSource` → step client): two sources forced
  onto one URL are rejected; a cached asset is refused across job sources. During development,
  deleting the single wiring line left the rest of the suite green — these tests exist so that exact
  defect can never ship silently: with the wiring removed they fail (verified both directions).
- Non-regression, matching what you verified on #172: three jobs against the same file still share
  one fetch (asserted at fetch-count level); unscoped callers (interactive viewer, docs hero) are
  unaffected; abort/in-flight-dedup behaviour unchanged.
- Full cadjs suite green. The one pre-existing Windows-only failure
  (`test_timestamp_output_path_preserves_extension`) is unchanged and untouched.

## Bundle

`skills/cad/scripts/snapshot/runtime/snapshot-render.js` regenerated with the pinned toolchain
(esbuild 0.27.7 / three 0.160.0 / gifenc 1.0.3, `bundle-cad.sh`'s exact flags). Reproducibility
check: two independent installs of the pinned deps produced byte-identical output
(sha256 `908bfa70…`), and the same invocation reproduces the previously committed bundle
byte-for-byte from base sources.


earthtojake (approved):
Reviewed and tested end-to-end.

Verified the fix actually closes the bug through the real stack: forced a collision by pinning two distinct sources onto one render-asset URL, then ran the real driver → Chromium → bundled `snapshot-render.js`. On `develop` the second job silently emitted the first job's image byte-for-byte at exit 0; on this branch it raises `RenderAssetSourceScopeError` naming both sources and exits 1.

No regression: a 3-job batch across two sources renders correctly and same-source jobs still share one fetch (byte-identical outputs). cadjs suite 407/407 locally, `scripts/bundle/bundle.sh --check` clean, CI green.

Confirmed the tests pin the wiring — deleting the single `setRenderAssetSourceScope` line in `loadSource` turns 4 tests red. Also confirmed the fail-closed shape is safe: `resolved.inputPath` is set at the single construction site in `__main__.py`, and no in-repo caller passes a `resolved` packet without it (the viewer builds none, the docs hero passes none).

Two non-blocking notes for whenever it's convenient, neither a defect in what's here:

1. The STEP parameter sidecar channel isn't covered — `loadStepModuleDefinition` uses a bare dynamic `import(url)`, which is a page-lifetime URL-keyed cache reached from inside the newly-scoped block. Same hypothetical collision would silently serve the wrong sidecar there.
2. `activeAssetSourceScope` is a module global that `loadSource` awaits across, so it assumes jobs render sequentially (true today at `__main__.py:1150`). Parallel jobs would make the guard misfire.

Minor: `releaseAssetSourceScope` is fully redundant with the occupancy re-derivation in `assertAssetSourceScope` — neutering either alone leaves the suite green; only removing both turns it red. Fine as deliberate defense in depth.

Approving.

Issue #125, opened 2026-07-01
CADview打开崩溃


liankaibihai:
我在codex中发现cad viewer出现有概率崩溃导致codex崩溃，我目前无法稳定浮现该问题

liankaibihai:
目前发现只要时间够长似乎能稳定崩溃

earthtojake:
Thanks for the report. We found and fixed one confirmed crash vector (a single malformed request — e.g. an invalid `dir=` query value — could kill the whole viewer server process) on branch `claude/github-issue-125-review-b2zx6t`. But to confirm that's what you're hitting, we need a bit more detail:

1. **How is the crash triggered?** For example:
   - Does it crash when you *open* the viewer in Codex after the server has been running for a while?
   - Or does the server suddenly crash while the viewer is already open in the browser?
   - Or does it crash right at startup?
2. **Crash logs.** If Codex is running the viewer, the crash output (stack trace) should be in the Codex thread that launched it — please ask that crashed thread to show the viewer server's error output, and paste the stack trace here.
3. Roughly how long does the server run before it crashes, and does anything correlate (opening a file, switching directories, generating a model)?

---

感谢反馈。我们在分支 `claude/github-issue-125-review-b2zx6t` 上已经找到并修复了一个确认的崩溃路径（一个畸形请求——例如非法的 `dir=` 查询参数——就能杀死整个 viewer 服务进程）。但为了确认这就是你遇到的问题，需要更多信息：

1. **崩溃是如何触发的？** 例如：
   - 是服务器运行一段时间后，在 Codex 中*打开* viewer 时崩溃？
   - 还是 viewer 已经在浏览器中打开，服务器突然崩溃？
   - 还是启动时就崩溃？
2. **崩溃日志。** 如果是 Codex 启动的 viewer，崩溃输出（堆栈信息）应该在启动它的那个 Codex 会话里——请让那个崩溃的会话把 viewer 服务的错误输出打印出来，并粘贴到这里。
3. 服务器大约运行多久后崩溃？崩溃时是否有相关操作（打开文件、切换目录、生成模型等）？

---
_Generated by [Claude Code](https://claude.ai/code)_

earthtojake:
Closing due to inactivity, please let me know if this is still an issue in version 0.4 @liankaibihai

PR #192, opened 2026-08-05
fix(viewer): survive malformed percent-encoding in legacy CAD paths

### LINKED ISSUE #191: GET /__cad/<invalid-percent-encoding> crashes the entire CAD Viewer server process
## Summary

A single HTTP request with malformed percent-encoding in a `/__cad/...` path crashes the whole CAD Viewer backend with an uncaught `URIError`, killing the server process and any agent interaction that launched it. This is the same class of "request randomly crashes the Viewer" issue as #125, but a **separate, still-unpatched vector**: the `dir=` query fix on `claude/github-issue-125-review-b2zx6t` does not cover this one.

## Version / Environment

- Branch: `develop` (tested at `7fcdfab`)
- Node v24.18.0 (reproducible on any modern Node ESM runtime)
- OS: macOS (likely OS-independent)

## Affected code

- `viewer/src/server/httpHandlers.mjs:90` — `decodeURIComponent(requestUrl.pathname.slice("/__cad/".length))` inside `legacyCadAssetFileRef()`
- Called from `localAssetMiddleware` at `httpHandlers.mjs:586` **outside** any try/catch
- No `process.on('uncaughtException')` handler exists in `viewer/src/server/server.mjs`, so the throw escapes the request listener and terminates the Node process

> Note: all other `decodeURIComponent` call sites in the repo are already guarded — `httpHandlers.mjs:630` (`serveDistAsset`), `viewer/src/server/localAssetBackend.mjs:247`, and `viewer/src/shared/viewerConfig.mjs:67-68`. Only this site is unguarded.

## Steps to reproduce (minimal, no dependencies needed)

1. Clone the repo on `develop`:

   ```bash
   git clone --branch develop https://github.com/earthtojake/text-to-cad.git
   cd text-to-cad/viewer
   ```

2. Run the following replication script (imports the real `createLocalAssetMiddleware` with a stub backend):

   ```bash
   node --input-type=module -e '
   import http from "node:http";
   import { createLocalAssetMiddleware } from "./src/server/httpHandlers.mjs";
   const backend = {
     assetPathForFileRef: () => "/tmp/something.step",
     contentTypeForPath: () => "application/octet-stream",
   };
   const mw = createLocalAssetMiddleware({ backend, rootDir: "/tmp" });
   const server = http.createServer((req, res) => mw(req, res, () => { res.statusCode = 404; res.end("nf"); }));
   server.listen(0, "127.0.0.1", () => {
     const p = server.address().port;
     console.log("server up on", p);
     http.get(`http://127.0.0.1:${p}/__cad/foo%zz.step`, () => process.exit(0));
   });
   '
   ```

3. Observe the process dies with `URIError: URI malformed` and a nonzero exit code.

## Steps to reproduce (full server)

1. Clone and install:

   ```bash
   git clone --branch develop https://github.com/earthtojake/text-to-cad.git
   npm install --prefix viewer
   ```

2. Start the server:

   ```bash
   node src/server/server.mjs --host 127.0.0.1 --port 4178 --dir /tmp --json
   ```

3. Send a request with invalid percent-encoding:

   ```bash
   curl "http://127.0.0.1:4178/__cad/foo%zz.step"
   ```

   Note that ``%zz` is not a valid URL escape sequence.

4. The server process exits with `URIError: URI malformed` printed to stderr and exit code 1. Subsequent requests fail because the server is gone.

## Expected behavior

A malformed request should be rejected gracefully (e.g. 400/404 response) and the server must keep running.

## Actual behavior

Uncaught exception kills the whole process.

## Stack trace

```
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at legacyCadAssetFileRef (viewer/src/server/httpHandlers.mjs:90)
    at localAssetMiddleware (viewer/src/server/httpHandlers.mjs:586)
    at Server.<anonymous>
    at Server.emit (node:events:509:28)
```

## Why `new URL` doesn't protect us

- `new URL(req.url, base).pathname` returns the **raw** percent-encoded string unchanged (`%zz` is left as-is).
- `decodeURIComponent("%zz")` then throws a `URIError: URI malformed`, which is currently uncaught.

## Suggested fix

Wrap the `decodeURIComponent` call in a try/catch and treat an unparseable path as "not a CAD asset" (fall through to `next()`), mirroring how the other malformed-request paths are already handled. A regression test should be added to `viewer/src/server/httpHandlers.test.mjs` asserting that a malformed `/__cad/` pathname yields a graceful response and does not crash the server.

## Related

- **#125 "CADview打开崩溃"** (open, reopened 2026-07-29) — https://github.com/earthtojake/text-to-cad/issues/125
  - Reporter: CAD Viewer inside Codex crashes intermittently and takes down the Codex session ("有概率崩溃导致codex崩溃"); not stably reproducible at first, later noted to crash reliably given enough runtime.
  - Maintainer confirmed and fixed one vector on `claude/github-issue-125-review-b2zx6t`: a malformed `dir=` query value could kill the whole viewer server process.
  - This report is a **second, distinct crash vector of the same class** (malformed request kills the process), still unpatched on `develop`.%

warun7:
Fixes #191.

## Problem

A single request to a `/__cad/` path containing invalid percent-encoding
(e.g. `GET /__cad/foo%zz.step`) crashed the entire CAD Viewer server. In
`viewer/src/server/httpHandlers.mjs:90`, `legacyCadAssetFileRef()` called
`decodeURIComponent()` on the raw pathname outside any try/catch, and the
resulting uncaught `URIError` escaped the request listener and terminated the
Node process.

All other `decodeURIComponent` call sites in the repo are already guarded
(`serveDistAsset` at httpHandlers.mjs:630, `localAssetBackend.mjs:247`,
`viewerConfig.mjs:67-68`); this was the only exposed one.

## Repro

```bash
node --input-type=module -e '
import http from "node:http";
import { createLocalAssetMiddleware } from "./viewer/src/server/httpHandlers.mjs";
const backend = { filePath: "x", assetPathForFileRef: () => "/tmp/a.step", contentTypeForPath: () => "app/octet-stream" };
const mw = createLocalAssetMiddleware({ backend, rootDir: "/tmp" });
const server = http.createServer((req, res) => mw(req, res, () => { res.statusCode = 404; res.end("nf"); }));
server.listen(0, "127.0.0.1", () => {
  const p = server.address().port;
  http.get(`http://127.0.0.1:${p}/__cad/foo%zz.step`, () => process.exit(0));
});
'
```

Before: `URIError: URI malformed` → process exits code 1.
After: malformed paths fall through to `next()`; the server stays up.

## Changes

- `viewer/src/server/httpHandlers.mjs` — guard `decodeURIComponent` in `legacyCadAssetFileRef`; return `""` on malformed input so the middleware falls through gracefully.
- `viewer/src/server/httpHandlers.test.mjs` — regression test asserting a malformed `/__cad/` pathname does not call the asset backend and falls through to the next middleware.

## Verification

- `node --test 'src/**/*.test.*js' 'scripts/**/*.test.*js'` in `viewer/` (Node 24): 401 tests, 0 failures
- `scripts/dev/setup-symlinks.sh --check`: development symlink layout is valid
- `git diff --check`: clean

Note: `npm --prefix viewer run test` fails on Node 24 with
`--experimental-default-type=module` (removed in newer Node); it expects Node 22.

### DIFF
--- a/viewer/src/server/httpHandlers.mjs
+++ b/viewer/src/server/httpHandlers.mjs
@@ -87,7 +87,12 @@ function legacyCadAssetFileRef(requestUrl, req) {
   if (!requestUrl.pathname.startsWith("/__cad/") || requestUrl.pathname === "/__cad/asset") {
     return "";
   }
-  const relativePath = decodeURIComponent(requestUrl.pathname.slice("/__cad/".length));
+  let relativePath = "";
+  try {
+    relativePath = decodeURIComponent(requestUrl.pathname.slice("/__cad/".length));
+  } catch {
+    return "";
+  }
   if (!relativePath || !path.extname(relativePath)) {
     return "";
   }
--- a/viewer/src/server/httpHandlers.test.mjs
+++ b/viewer/src/server/httpHandlers.test.mjs
@@ -375,6 +375,31 @@ test("local asset middleware resolves legacy URDF mesh URLs from referrer file",
 });
 
 
+test("local asset middleware survives malformed percent-encoding in legacy CAD paths", async () => {
+  const middleware = createLocalAssetMiddleware({
+    backend: {
+      assetPathForFileRef: () => {
+        assert.fail("assetPathForFileRef must not be called for a malformed path");
+      },
+      contentTypeForPath: () => "model/stl",
+    },
+  });
+  const req = {
+    method: "GET",
+    url: "/__cad/foo%zz.step",
+    headers: {},
+  };
+  const res = createResponse();
+  let nextCalled = false;
+
+  middleware(req, res, () => {
+    nextCalled = true;
+  });
+
+  assert.equal(nextCalled, true, "malformed legacy CAD path should fall through to the next middleware");
+});
+
+
 test("CAD Viewer API middleware reveals file assets with POST reveal route", async () => {
   const calls = [];
   const middleware = createCadViewerApiMiddleware({


earthtojake (approved):
Verified end-to-end. Correct, minimal fix for a real remote-crash bug.

## Bug confirmed on unpatched `develop`

Full server (`server.mjs --port 4193 --dir /tmp`), single request:

```
healthy request       → 404 (server up)
GET /__cad/foo%zz.step → status=000
process alive?        → DEAD
next request          → connection failed
URIError: URI malformed ... at runMiddleware (server.mjs:190) → exit 1
```

## Fix confirmed on this branch

| Test | Result |
|---|---|
| `GET /__cad/foo%zz.step` | `400 Bad request`, process alive |
| 30 malformed requests hammered | all `400`, still serving |
| 56-request fuzz sweep — truncated `%E0%A4%A`, overlong `%C0%80`, `%00`, malformed `dir=`/`file=`/`referer`, traversal | all graceful; baseline died on request #1 |

Falling through to `serveDistAsset`'s existing guard yields `400 Bad request`, matching the expected behavior in #191.

## Test genuinely catches the regression

Mutation test — guard reverted, new test kept:

```
not ok 1 - local asset middleware survives malformed percent-encoding in legacy CAD paths
  name: 'URIError'
```

## Fix is complete for this class

Audited all 7 `decodeURIComponent` sites in `viewer/src` + `packages/cadjs/src`. The other request-reachable ones (`httpHandlers.mjs:635`, `localAssetBackend.mjs:247`) were already guarded, and `requestRefererUrl`'s `new URL` is guarded. This was the last exposed one, as the PR states.

## Checks

- `scripts/test/test-js.sh` → exit 0: cadjs 407/407, implicitjs 56/56, viewer 401/401 (confirms the claimed 401 on Node 22)
- `scripts/dev/setup-symlinks.sh --check` → valid
- `git diff --check` → clean
- No generated runtime to refresh: `skills/cad-viewer/scripts/viewer` symlinks to `viewer/`, so the edited file is the only source

## Two minor notes (non-blocking, not worth another round)

1. The `assert.fail` inside the test's `assetPathForFileRef` stub is dead code — `localAssetMiddleware` wraps that call in its own try/catch and calls `next()` on any non-403 error, so the `AssertionError` is swallowed (probed directly: `assetPathForFileRef called: true, next() called: true`, nothing thrown). The test still works via the `URIError` path, but the "must not be called" guarantee isn't enforced; a boolean flag asserted afterwards would make it real. Same block: `contentTypeForPath` is unused.

2. The amplifier is untouched. `runMiddleware` (`server.mjs:184`) calls each middleware with no try/catch and there is no `uncaughtException`/`unhandledRejection` handler — which is why a one-line decode bug becomes a process kill. This is the second vector of this class patched at the call site (#125 was the first). Out of scope here; tracking a top-level guard separately.

PR #194, opened 2026-08-05
fix(viewer): contain handler errors instead of killing the server

earthtojake:
## Problem

The CAD Viewer backend had no top-level protection against exceptions thrown from HTTP request handlers. `runMiddleware` invoked each middleware with no `try`/`catch`, and `http.createServer((req, res) => runMiddleware(0, req, res))` meant any synchronous throw escaped to the request listener and terminated the process — taking down any agent session that launched the Viewer.

This is a recurring class, not a one-off:

- #125: a malformed `dir=` query value killed the server process.
- #191 / #192: `GET /__cad/foo%zz.step` killed the server via an unguarded `decodeURIComponent` in `legacyCadAssetFileRef`.

Both were fixed at the individual call site. The amplifier was untouched, and there were no `unhandledRejection` / `uncaughtException` handlers anywhere in the backend.

## Change

The middleware chain moves out of `server.mjs` into a new `viewer/src/server/requestPipeline.mjs` so it is testable in isolation, and every `middleware(req, res, next)` call is contained:

- **Synchronous throws** are caught per-hop. `next()` recurses inside the parent's `try`, but each level has its own catch, so a downstream throw is handled at its own frame and never propagates outward.
- **Async rejections** are handled via `Promise.resolve(result).catch(fail)` when a middleware returns a thenable. Several middlewares are `async` and their promises were previously dropped entirely.
- **Responses**: 500 `Internal Server Error` when `!res.headersSent`; `res.destroy()` when headers are already on the wire; a nested `try`/`catch` in case sending the 500 itself fails. A latch means only the first failure per request responds.
- **404 fallthrough** behavior is unchanged.

`installProcessErrorGuards` adds last-resort `unhandledRejection` and `uncaughtException` handlers that log and keep the process alive.

`uncaughtException` was a deliberate call rather than an oversight: async throws from fs/stream callbacks escape the per-middleware `try` entirely, which is exactly the amplifier class in scope here. To offset the bug-masking risk, the pipeline tracks in-flight requests and both guards include them in the log line:

```
CAD Viewer backend uncaught exception: Error: ... [in-flight: GET /__cad/foo%zz.step]
```

## Verification

- 8 new tests in `viewer/src/server/requestPipeline.test.mjs`, all against a real `http.createServer`: sync throw → 500 and a follow-up request still returns 200; async rejection → 500 and still 200; failure after headers sent → response destroyed, server survives; 404 fallthrough; single-report latch; in-flight tracking; both process guards log without exiting.
- `npm --prefix viewer run test` → 409/409 pass.
- `scripts/test/test-js.sh` → cadjs 407, implicitjs 56, viewer 409, all pass (Node 22).
- **Live end-to-end**: on the pre-#192 tree, `GET /__cad/foo%zz.step` against the real server returned **500** with the `URIError: URI malformed` stack logged and the server stayed responsive — where that same request previously exited the process. After rebasing onto current `develop`, the request returns 400 from #192's call-site guard, so the two layers compose as intended.
- The pre-commit hook ran the full bundle check, including the esbuild bundle of the backend with the new module.

Nothing to regenerate: `bundle-cad-viewer.sh` esbuilds `src/server/server.mjs`, so the new sibling module is picked up automatically.

### DIFF
--- a/viewer/README.md
+++ b/viewer/README.md
@@ -90,7 +90,9 @@ review links.
 - `src/client/ui/`: viewer-owned browser utilities such as clipboard, color
   scheme, class merging, and DOM helpers.
 - `src/server/`: local and hosted backend adapters, HTTP middleware, and the
-  local production server.
+  local production server. `src/server/requestPipeline.mjs` runs the middleware
+  chain and contains handler failures: an unexpected throw or rejection becomes
+  a 500 instead of killing the server process.
 - `api/cad/`: Vercel serverless shims for `/__cad/*` routes.
 - `scripts/`: developer and runtime launchers, plus the test runner.
 - `docs/`: workflow reference docs for backend storage, browser persistence,
--- a/viewer/src/server/requestPipeline.mjs
+++ b/viewer/src/server/requestPipeline.mjs
@@ -1,0 +1,130 @@
+const DEFAULT_LOG = (message) => process.stderr.write(`${message}\n`);
+
+function describeError(error) {
+  if (error instanceof Error) {
+    return error.stack || `${error.name}: ${error.message}`;
+  }
+  return String(error);
+}
+
+function describeRequest(req) {
+  const method = String(req?.method || "?");
+  const url = String(req?.url || "");
+  return url ? `${method} ${url}` : method;
+}
+
+/**
+ * Turn an unexpected handler failure into a 500 instead of an unhandled throw.
+ * Once headers are on the wire a status code is no longer negotiable, so the
+ * response is destroyed to signal the truncation to the client.
+ */
+export function respondWithInternalError(error, req, res, { log = DEFAULT_LOG } = {}) {
+  log(`CAD Viewer backend request failed (${describeRequest(req)}): ${describeError(error)}`);
+  if (!res || res.writableEnded) {
+    return;
+  }
+  try {
+    if (res.headersSent) {
+      res.destroy();
+      return;
+    }
+    res.statusCode = 500;
+    res.setHeader("Content-Type", "text/plain; charset=utf-8");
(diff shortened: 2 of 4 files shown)


earthtojake:
Closing without merging: `release/0.4.0` is about to merge into `develop`, and that merge deletes the Node backend this PR hardens.

`release/0.4.0` replaced `viewer/src/server/*.mjs` with the Python backend (`viewer/server_py/`), so `runMiddleware` and the request listener this PR rewrites will not exist on `develop` after the merge. The bug class is also already contained there, at two levels:

- `Handler.do_GET` / `do_POST` wrap dispatch in `except ForbiddenAssetError -> 403` / `except Exception -> 400`.
- Anything escaping that only reaches `socketserver.BaseServer.handle_error`, which logs a traceback and closes the connection — the process survives. Verified with a `ThreadingHTTPServer` whose handler raises unconditionally: the failing request drops, the next one returns 200, the process stays up.

The #191 shape is already guarded specifically on that branch too (`_legacy_cad_asset` uses `_strict_decode_uri_component` with `except ValueError`), and Python's `urlsplit`/`parse_qs` do not raise on `%zz` the way JS `decodeURIComponent` does. That asymmetry is why #125 and #191 were Node-backend-only process kills.

Two small residuals remain on the Python side, neither fatal and neither addressed here:

1. `path, q = self._query()` sits outside the `try` in `do_GET`/`do_POST`, so a failure there drops the connection instead of returning a status.
2. The blanket `except Exception -> 400` reports server-side bugs as client errors; `500` would classify them better.

Worth folding into other `release/0.4.0` work rather than a dedicated backport. The branch `claude/vibrant-einstein-f7c35d` is left in place if any of this needs revisiting.

PR #201, opened 2026-08-10
Fix CAD Viewer Node 24 reproducibility

823608044-droid:
## What changed

- remove obsolete `--experimental-default-type=module` flags from Node test runners
- preserve symlink-aware Viewer resolution on current Node releases
- declare npm and pnpm install-script policies for esbuild/fsevents
- make the Viewer bundle choose the package manager from the committed lockfile, with an explicit override
- add focused policy regression coverage

## Why

The previous bundle path could select pnpm merely because it existed on the machine even when the repository committed `package-lock.json`. That made identical source revisions produce machine-dependent Viewer installs. Current Node releases also reject the obsolete test-runner flag path used by the affected scripts.

## Impact

CAD Viewer tests and bundled runtime installation are reproducible on Node 24 without changing geometry behavior or the installed plugin until a release is cut.

## Validation

- `bash scripts/test/test.sh`
- Viewer: 401/401 tests
- focused global policy: 8/8 tests
- Python CAD suite included in the full script
- `git diff --check`

### DIFF
--- a/packages/cadjs/scripts/run-tests.mjs
+++ b/packages/cadjs/scripts/run-tests.mjs
@@ -29,7 +29,6 @@ if (!tests.length) {
 
 const result = spawnSync(process.execPath, [
   "--test",
-  "--experimental-default-type=module",
   ...tests,
 ], {
   cwd: packageRoot,
--- a/packages/implicitjs/package.json
+++ b/packages/implicitjs/package.json
@@ -2,6 +2,9 @@
   "name": "implicitjs",
   "version": "0.1.0",
   "type": "module",
+  "allowScripts": {
+    "fsevents": false
+  },
   "description": "Standalone browser-native implicit CAD runtime, renderer, mesh sampler, and exporters.",
   "exports": {
     ".": {
--- a/packages/implicitjs/scripts/run-tests.mjs
+++ b/packages/implicitjs/scripts/run-tests.mjs
@@ -32,7 +32,6 @@ if (!tests.length) {
 
 const result = spawnSync(process.execPath, [
   "--test",
-  "--experimental-default-type=module",
   ...tests,
 ], {
   cwd: packageRoot,
--- a/scripts/bundle/skills/bundle-cad-viewer.sh
+++ b/scripts/bundle/skills/bundle-cad-viewer.sh
@@ -100,6 +100,14 @@ resolve_viewer_package_manager() {
     echo "$VIEWER_PACKAGE_MANAGER"
     return
   fi
+  if [ -f "$VIEWER_DIR/package-lock.json" ]; then
+    echo "npm"
+    return
+  fi
+  if [ -f "$VIEWER_DIR/pnpm-lock.yaml" ]; then
+    echo "pnpm"
+    return
+  fi
   if command -v pnpm >/dev/null 2>&1; then
     echo "pnpm"
     return
--- a/tests/python/global/test_documented_viewer_commands.py
+++ b/tests/python/global/test_documented_viewer_commands.py
@@ -10,7 +10,10 @@ generates.
 from __future__ import annotations
 
 import json
+import os
 import re
+import subprocess
+import tempfile
 import unittest
 from pathlib import Path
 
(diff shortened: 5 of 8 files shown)


earthtojake:
Hi, thanks for this, can you point at the latest develop branch and verify that it's still an issue? I just published a big upgrade with version 0.4

earthtojake:
Thanks for your patience — I rebuilt this against current `develop` (`58ff5704`) so you don't have to guess what survived the 0.4 upgrade. Answering my own question above: **yes, the core issue still reproduces**, and it's the most important thing in this PR.

## The flag issue is real and still there

All three runners still carry `--experimental-default-type=module` on `develop`, and Node 24 rejects it outright. Installing 24.19.0 and running against current `develop`:

```
npm --prefix packages/cadjs test        → bad option: --experimental-default-type=module   (exit 9)
npm --prefix packages/implicitjs test   → bad option: --experimental-default-type=module   (exit 9)
npm --prefix viewer run test            → bad option: --experimental-default-type=module   (exit 9)
```

The runner never starts. And 24.19.0 is the current Latest LTS, so `nvm install --lts` is enough to hit it — worth stating plainly, because our CI pins Node 22 and will never catch this.

I confirmed your fix is sufficient, not just necessary. Node 24 with the flag deleted matches the Node 22 baseline exactly:

| suite | Node 22 today | Node 24, flag deleted |
| --- | --- | --- |
| cadjs | 387 / 370 pass / 17 fail | 387 / 370 pass / 17 fail |
| implicitjs | 64 / 55 pass / 9 fail | 64 / 55 pass / 9 fail |
| viewer | 299 pass | 299 / 299 pass |

(Those residual failures are unhydrated `node_modules` in my worktree, identical on both versions — not yours and not Node 24's.)

Scope, for the record: the flag is test-only. End-user artifact generation passes 12/12 on Node 24 and `vite build` is clean, so no shipped runtime is affected.

## The lockfile ordering is also a real bug

`resolve_viewer_package_manager` picks pnpm whenever pnpm merely exists, but we commit `viewer/package-lock.json` and no `pnpm-lock.yaml`. I hit this myself earlier this week without recognising it — a `bundle.sh` run used pnpm on an npm-locked tree. Your lockfile-first ordering is the right fix.

## You were right about `allowBuilds` and I was wrong

I assumed this was cargo-culted and tested it against a control with a bogus key. Baseline and bogus key both print `Ignored build scripts: esbuild@0.27.7`; with `allowBuilds` the postinstall runs. It's a genuine pnpm key and it matters here, since esbuild without its postinstall has no binary — exactly what the bundler's `node_modules/.pnpm` esbuild fallback exists to work around. Apologies for the doubt; keep it.

## What to change

Two things to **drop**:

- `--preserve-symlinks` isn't needed — 299/299 with or without it. Deleting the obsolete flag is enough.
- `allowScripts` in `package.json` is inert; there's no lavamoat in the repo and npm doesn't read that field. The `pnpm-workspace.yaml` half is the part that works.

One thing that needs **rewriting**: the ~84 lines in `tests/python/global/test_documented_viewer_commands.py`. That file no longer exists on `develop`, and its assertions reference `viewer/src/server/serverArgs.mjs` — the entire Node backend, replaced by a Python one in 0.4. That's also why GitHub shows this as conflicting: it's a delete/modify conflict, not a textual one.

## Suggested path

Rebasing this is likely more painful than re-cutting it, since most of the added lines target a runtime that no longer exists. If it were me I'd open a fresh PR against `develop` with just the substantive ~12 lines: delete the flag from the three runners, add the lockfile-first branches to the bundler, add `viewer/pnpm-workspace.yaml`, and write new policy tests against the current Python backend.

Happy either way — rebase this one or open a new one, whichever you prefer. If you'd rather not spend more time on it, say so and we'll land it ourselves with credit to you. Just let us know within a week or two so it doesn't go stale again.

earthtojake:
Landed the still-live findings from this report in #234 — thank you for the diagnosis, it was accurate about the mechanism and the two independent causes.

Re-verified each item against `develop` first, since this branch predates the 0.4.x viewer and can't be rebased:

- **cadjs + viewer runners** — already fixed on `develop`; both suites run on Node 18→24 today. #234 adds a policy test so the flag can't come back.
- **implicitjs** — still broken, and worse than described. The flag sat behind `nodeMajor < 22`, which is wrong at both ends: Node 18 has no such flag (`bad option`, rc=9, zero tests run), and on Node 20 the flag reaches only the runner process, so the two tests that spawn the export CLI failed anyway (349/2). The real constraint is that the fixtures are bare `.js` files with `export default`, so #234 states the Node 22 floor instead of passing a flag that breaks one version and half-fixes another. Same wall applies to real `.implicit.js` models, so the loader now blames the interpreter rather than the user's model.
- **package-manager selection** — still broken, fixed as you proposed: the committed lockfile decides, `package-lock.json` first so a stray local `pnpm install` can't flip a release either. Your extract-and-run test harness is in #234 largely as you wrote it.
- **`--preserve-symlinks`** — not landed. The viewer suite passes on 18/20/22/24 without it, and it changes module resolution for the symlinked dev layout.
- **`allowScripts` / `pnpm-workspace.yaml allowBuilds`** — not landed. Nothing reads either key here: `allowScripts` is `@lavamoat/allow-scripts` (not a dependency) and pnpm's key is `onlyBuiltDependencies`.

Closing in favour of #234.

earthtojake:
#234 is merged into `develop`, so closing this as superseded — the still-live findings landed there with credit to you, including the `pnpm-workspace.yaml` `allowBuilds` policy after I re-ran your case against a bogus-key control and confirmed you had the right key. Thanks again for the diagnosis.

PR #224, opened 2026-08-12
cadgen: bundle the last unbundled Node entry, and pin that STEP never needs Node

earthtojake:
First step of making cadgen a self-contained, installable package: remove the last unbundled Node entry, and pin the guarantee the rest of the work has to preserve.

## Why this is the first step

Skills vendor 66 files of cadgen each, purely so `node_package_root()`'s `Path(__file__).parents[4]` lands next to a `packages/cadjs/bin`. Removing that archaeology is the goal — but one Node entry could not be bundled, and an unbundled entry needs its whole dependency **graph** at runtime, which is why implicit-cad also vendored all of implicitjs and why the child needs a resolve hook and `NODE_PATH`.

## 1. STEP must never need Node — now enforced

STEP is the core build path and is pure OCP/build123d. I verified it builds with `node` entirely absent from `PATH`, then pinned it.

A static import check would have been wrong: `cadgen.step_artifact` **does** import the Node bridge transitively (`step_artifact → component_package → generation → drawing_package → node_runtime`). That edge is inert — `node_runtime` is stdlib-only and `cad_node_executable()` resolves at call time — so the test asserts *behaviour*: point `CADGEN_NODE` at a non-executable, build a STEP package, and require success.

Verified both directions — passes today, fails with `NodeUnavailable` when a `cad_node_executable()` call is injected into `generate_step_targets`.

## 2. The export CLI is now a builder like the others

`implicitjs/scripts/export.mjs` moves behind a thin wrapper at `packages/cadjs/bin/implicit-export.mjs`, added to implicit-cad's `BUILDER_ENTRIES`, esbuilt into one self-contained **82 KB** file. `implicit_export_cli()` resolves through `node_builder_script()` like everything else. The CLI implementation stays in implicitjs and is still its own `npm run export` — one implementation, two entry points.

## 3. implicit-cad stops vendoring implicitjs — **−59 published files**

Nothing reads the tree any more. Also collapses a fork in the bundle script: the check path had a separate "runtime is a symlink, only verify builders" branch that existed solely because this output was a copy in production and a link in development.

## A silent failure this turned up

The first bundle came out as a **20-byte shebang**. implicitjs's `sideEffects` listed only `implicitHeadlessRenderEntry.js`, so esbuild tree-shook away `import "implicitjs/cli/export"` — a script whose entire purpose *is* its side effect. esbuild notes it and exits 0; the file existed, so every check passed. The only symptom would have been a builder that runs and does nothing.

Fixed the metadata, and added a guard: `bundle_node_builders` rejects a builder that emits no code. A raw size floor was my first attempt and it was wrong — `implicitClosureHooks.mjs` is a legitimate 370 bytes — so it strips the shebang and checks what's left. Confirmed it fires on the exact regression and passes all four real builders.

## Verification

- Real model (`boys-surface-sculpture`) exports to STL through **both** `cadgen.export_implicit_model` and the skill's own `export.mjs`, including after the vendored tree was deleted
- `scripts/test/test.sh` exit 0 — 568 cadjs / 351 implicitjs / 346 viewer, 10 Python suites
- `bundle.sh --check`, `setup-symlinks.sh --check`, `check-version.sh` all clean

## Follow-up, not in this PR

**Phase 2 (injection):** delete `node_package_root()`'s `parents[4]` and have callers supply the builder directory, the way `run_snapshot_cli(runtime_dir=…)` already does. Only three call sites inside cadgen, but every skill entry point and the viewer backend must then supply the location, and a missed one breaks DXF/implicit builds at runtime in a way unit tests won't catch — so it wants a careful pass and per-skill end-to-end checks.

**Phase 3 (un-vendor cadgen):** ~396 files leave `skills/`. Depends on Phase 2, and makes `pip install -r requirements.txt` a hard prerequisite — worth its own PR since no SKILL.md documents that step today.


earthtojake:
Closing unmerged, per review: this was written as phase 1 of a per-skill layout that has since been superseded by the cadgen mega-package decision. The branch stays as a parts bin — `cadgen-unification-plan.md` on `claude/cadgen-unification-plan` (§0.1) lists the four pieces to carry forward (STEP-no-Node test, implicit-export wrapper + sideEffects fix, no-code bundler guard, builder-name resolution) and the two to skip (per-skill bundle destination, early implicitjs un-vendor).

PR #225, opened 2026-08-13
Phase A: cadgen carries the JavaScript it executes

earthtojake:
First of four phases making `cadgen` a self-contained distribution. **No behaviour changes** — this moves where cadgen finds the JS it already runs.

## The problem

cadgen spawns Node builders (DXF, implicit) and drives a browser bundle (snapshot), but shipped neither. It located them by walking up from its own `__file__`:

```python
return Path(__file__).resolve().parents[4]   # a sibling packages/ dir
```

That one line is why every skill vendors a **whole copy of cadgen** — not for the Python, but so the walk lands next to a `packages/cadjs/bin`.

## The change

**`cadgen/assets.py`** — one resolution order, call-time only: explicit env → live sources of a source checkout → the copy shipped in the wheel. Dev-source-first is deliberate, so builder JS stays editable without a rebundle. Nothing runs at import: `pip install cadgen` must work with no Node, and the viewer's long-lived server must import light.

**`cadgen/_runtime/`** — package data holding `node/` (5 esbuilt builders), `browser/` (snapshot bundle), `moveit2/`, and `viewer/` (the SPA). Built by the new `bundle-cadgen-runtime.sh`.

**The implicit export CLI is now a builder.** It was the only unbundled Node entry, and an unbundled entry needs its whole dependency graph at runtime — the reason implicit-cad vendors all of implicitjs. A thin wrapper in `packages/cadjs/bin` bundles it to 82 KB self-contained; the CLI still lives in implicitjs and remains its own `npm run export`.

**`check-wheel-contents.sh`** builds the wheel and asserts the assets are inside, wired into `test.yml` and into `release.yml` ahead of the PyPI upload. A package-data glob that stops matching yields a wheel that imports fine and fails on a user's machine — the same silent-failure shape as the bug below.

## A silent failure this turned up

The export wrapper first bundled to a **20-byte shebang**. implicitjs's `sideEffects` didn't name the script, so esbuild tree-shook away a side-effect-only import and exited 0. Fixed the metadata, and `bundle_node_builders` now rejects a builder emitting no code — stripping the shebang rather than using a size floor, since `implicitClosureHooks.mjs` is a legitimate 370 bytes.

## Also: the STEP tripwire

STEP needs no Node today, and one stray call in the shared generation path would break that silently in Node-less environments. `test_step_needs_no_node.py` pins it behaviourally (a static check would be misleading — `step_artifact` *does* import the Node bridge transitively, but the edge is inert). Verified it passes, and fails with `NodeUnavailable` when such a call is injected into `generate_step_targets`.

## Measurements

| | |
|---|---|
| Wheel | **4.7 MB** compressed |
| `_runtime` entries | 63 with the viewer, 25 without |
| Builders | dxf 634 KB · implicit-artifact 190 KB · implicit-export 82 KB · hooks 370 B · worker 40 KB |
| Viewer dist | 16 MB on disk (12 MB sourcemaps), gitignored, still ships |

Licensing: shipping three/meshoptimizer/gifenc in a wheel is redistribution, so `--legal-comments` moves `none` → `eof` and each runtime dir carries `THIRD_PARTY_LICENSES.txt`. That restamps the still-existing per-skill `dxf-artifact.mjs` (the only bundle inlining `three`); Phase C deletes those copies.

## Verification

- `scripts/test/test.sh` exit 0 — 568 cadjs / 351 implicitjs / 346 viewer, 10 Python suites
- `bundle.sh --check`, `setup-symlinks.sh --check` clean
- DXF gen bakes `preview.glb` and an implicit model exports to STL with `CADGEN_NODE_BUILDERS_DIR` pointed at the packaged runtime (i.e. the installed path)
- Wheel-contents check green with and without the viewer dist required

## Next

B (viewer backend into cadgen, retire the mirror) and C (skills stop vendoring) both depend on this and are independent of each other. `cadgen-unification-plan.md` rides along and is deleted in Phase D.


earthtojake:

PR #226, opened 2026-08-13
Phase B: move the CAD Viewer backend into cadgen, retire the mirror

earthtojake:
Stacked on #225 (Phase A). **Base is `claude/cadgen-phase-a`**, so this diff shows only Phase B; retarget to `develop` once #225 merges — CI only runs on PRs to `develop`, so this shows no checks until then.

## Why

The viewer's Python backend made **11 imports from cadgen's `_internal`** against 5 public ones. It was never a separate component — it shared cadgen's locks, freshness rules and coordination internals across a package boundary that existed only on disk, kept alive by a dual-import shim and a PYTHONPATH search so the two halves could find each other.

## What

- `viewer/server_py/` → `cadgen/viewer/` (16 modules, mostly R100 renames); its tests → `tests/python/packages/cadgen/viewer/`. Shim and path search deleted.
- **`pip install cadgen` now gets a working Viewer with no clone.** New `cadgen` console script dispatches to subcommands — only `viewer` so far, Phase C registers the rest — with `python -m cadgen.viewer` as the equivalent launcher. The built client is served from the wheel's `_runtime/viewer`. Dispatch imports lazily, so `--help` never pays for the OCP import.
- **Mirror retired.** `earthtojake/cad-viewer` was a verbatim copy kept in sync by a release job. With the backend inside cadgen the viewer is no longer a self-contained tree to copy, so the rule it enforced no longer holds. Deletes the sync script, its workflow, the release job, the self-containment test, and the `viewer/packages/cadgen` + `skills/cad-viewer/scripts/viewer` symlinks.

## Fixes found while gating this

| | |
|---|---|
| `worker_client` PYTHONPATH | Pointed at the cadgen package dir, not its parent — cannot import `cadgen.viewer`, and leaks every submodule as a bare top-level name (`import viewer`). Invisible wherever cadgen is installed, so it is pinned by a test that runs the import for real. |
| `cad-python.mjs` candidate order | Led with `viewer/packages/cadgen/src`, which walks **up** — out of a git worktree and into its parent checkout, binding the dev backend to another branch's cadgen. Nearest-first now, plus a test that every entry stays inside the checkout. |
| Release publish guard | Refused to publish unless `skills/cad-viewer/scripts/viewer` existed — a path this PR deletes. The wheel-contents check already enforces the real invariant (`CADGEN_REQUIRE_VIEWER_DIST=1`) before the trim. |
| e2e sweep fixtures | Named five paths that exist on **no** branch, so the sweep failed on fixture 1 for reasons unrelated to any change under test. Paths corrected; a preflight now reports missing/unhydrated fixtures as a broken gate rather than a render failure. |

## Gates

- Full suite green — 568 / 351 / 345 JS, 11 Python suites.
- `bundle.sh --check` with the viewer required.
- **Installed-wheel smoke:** wheel built, installed into a scratch venv, and from an *empty* cwd `cadgen viewer` serves the SPA and a 444-entry catalog. All three asset resolvers hit the packaged runtime.
- **e2e sweep 7/7** against that installed wheel.

Under the Vite **dev** server, implicit renders blank via a three.js `checkMaterialsReady` error inside a pre-bundled dep chunk. Reproduced with develop's own `implicitjs/package.json`, and no implicit source differs from develop — pre-existing and dev-only; the shipped bundle renders it fine. Worth a follow-up, not a blocker here.

## Your call, not done here

- Archive the `earthtojake/cad-viewer` repo (or push a final pointer README).
- Delete the now-unused `CAD_VIEWER_SYNC_TOKEN` secret.


earthtojake:

PR #227, opened 2026-08-13
Phase C: skills install cadgen instead of vendoring it

earthtojake:
Stacked on #226 (Phase B) -> #225 (Phase A). Base is `claude/cadgen-phase-b`, so this diff shows only Phase C; retarget down the stack as each merges. CI only runs on PRs to `develop`, so no checks show until then.

**120 files changed, +1,294 / -39,841.**

## Why

Six skills each carried a copy of the cadgen source tree, and two of them their own esbuilt Node builders. One runtime, duplicated six times, kept in step by six bundle scripts and six symlink-setup scripts. That duplication is the bulk of what this deletes.

## What

A skill names the `cadgen` **distribution** in `requirements.txt` -- unpinned on develop so the editable install satisfies it, pinned at publish, **with extras preserved** (`cadgen[snapshot]==X`, not `cadgen==X`; the bare form silently drops playwright from every published rendering skill). Its `scripts/<tool>` is a ~30-line entrypoint that forwards to cadgen and, when cadgen is missing, prints the pip command instead of a traceback.

Parsers moved **verbatim** into `cadgen/cli/`; the warm daemon into `cadgen/daemon/`. External interfaces are unchanged -- `scripts/gen --help` still prints `usage: scripts/gen`, `node scripts/export.mjs` still prints its own usage. `prog` became a parameter rather than a literal, so `cadgen step gen` names itself correctly while the skill entry keeps its own name.

The `cadgen` console script gains its remaining subcommands (`step`/`dxf`/`implicit` x `gen|artifact|export|inspect|snapshot`, plus `snapshot`, `viewer`, `daemon`), dispatched lazily so `--help` never pays for the OCP import.

## New gate: `scripts/test/test-installed.sh`

Wired into `test.yml` after `test.sh`. Builds the wheel, installs it into a scratch venv, and from an **empty directory** asserts: the wheel's own cadgen wins, all three asset resolvers land inside it, every subcommand dispatches, a real STEP builds, and `cadgen viewer` serves the SPA plus a catalog.

Every other check in this repo runs against the source tree, where the repo root is importable and every asset is one directory away. This one found three things nothing else could:

| | |
|---|---|
| **Module eviction ate numpy** | `_interpreter_roots()` derived third-party roots from `sys.prefix`/`sysconfig` alone, so a `site-packages` reached by `.pth`, `PYTHONPATH`, or a vendored bundle was classified as *model code* and evicted from `sys.modules` before a build. Fatal for a C extension, and it does not name the cause -- numpy reports a missing `dtypes` attribute. Any `site-packages` on `sys.path` now counts. |
| **Daemon keyed on a moved path** | Its socket and staleness token used a repo-relative path that resolves to the wrong directory post-move, and it watched a skill tree that no longer holds code it imports. Both now key on the cadgen package; the token pairs `cadgen.__version__` with the newest `.py` mtime -- the version alone cannot see an edit in a checkout, mtimes alone cannot see two wheels installed in turn. |
| **A bug I introduced** | I patched the wrong `build_parser()` call site and left a latent `NameError` in `inspect_command_result`. Caught here; that path is now exercised. |

## Policy tests updated, not weakened

- **Payload budget** now derives its own coverage from `json.dumps` usage instead of a hand-written list -- which immediately surfaced two parsers the list had never covered.
- **Builder bundles** assert the cadgen runtime, plus a new "no skill vendors a builder any more".
- **Pin script** gained the extras case.
- **`check-builds.sh`** sweeps `skills/` for symlinks directly, since the generated-outputs list no longer covers it. That invariant matters most here: Codex plugin installs drop symlinks silently.

## Gates

Full suite green (568/351/345 JS, 11 Python suites, 59 policy tests); docs; `bundle --check` with the viewer required; `check-builds`; all 13 skill CLIs smoke-tested plus a real DXF package build through the Node path; installed-mode gate green end to end.

[BOT] Generated with [Claude Code](https://claude.com/claude-code)


earthtojake:

PR #228, opened 2026-08-13
Phase D: docs catch up with the reorganization, plan file removed

earthtojake:
Stacked on #227 (Phase C) -> #226 -> #225. Base is `claude/cadgen-phase-c`; retarget down the stack as each merges.

Mechanical, but the stale text pointed at paths that no longer exist.

## What

- **`packages/cadgen/README.md`** is the PyPI front page. It still described a package whose "boundary is intentionally narrow", told readers to install a vendored copy from `./scripts/packages/cadgen`, and pointed at two symlinks that are gone. Rewritten around what the distribution actually is now: the Python runtime, the CLI, and the JavaScript it executes, installed by one pip command. It documents the two things cadgen deliberately does *not* carry -- Node on `PATH` and a snapshot browser -- and which commands need them, since that is the first thing a new install trips on.
- **The develop-layout guard** still refused a tracked `skills/cad-viewer/scripts/viewer/dist`. That path is gone; the equivalent mistake now is committing `cadgen/_runtime/viewer` (a 16 MB Vite output), so the guard watches that instead.
- Comment and doc references to `bundle-cad.sh`, `inspect_refs`, `cadgen_daemon` and the per-skill snapshot runtimes retargeted at where that code lives.
- Deletes `cadgen-unification-plan.md`, now executed end to end.

Extras (`[snapshot]`) and the viewer `THIRD_PARTY_LICENSES` landed in Phases C and A respectively, so nothing was left to do for those.

## Gates

Full suite, docs, `bundle --check` with the viewer required, `check-builds`, installed-mode, version, symlink layout -- all green.

## Your call

- **Build-test rehearsal** -- `Release` with `bump=none`, `target_branch=build-test`. Not dispatched; say the word and I will.
- **The real release** -- you pick the bump.
- Left over from Phase B: archive the `earthtojake/cad-viewer` repo and delete the `CAD_VIEWER_SYNC_TOKEN` secret.

[BOT] Generated with [Claude Code](https://claude.com/claude-code)


earthtojake:

PR #229, opened 2026-08-13
release 0.5.0

earthtojake:
`cadgen` becomes the single distribution: the Python runtime, every CLI parser, the CAD Viewer (backend **and** built client), the warm daemon, and the JavaScript it executes. `pip install cadgen` is the whole thing.

**25 commits + 4 develop merges. 397 files, +6,648 / -41,950.** Reviewable in order; each step was gated on its own.

Supersedes the stacked #225 / #226 / #227 / #228.

---

## Why

The viewer's Python backend made **11 imports from cadgen's `_internal`** against 5 public ones — never a separate component, just a package boundary that existed on disk. Meanwhile six skills each carried a copy of the cadgen source tree, kept in step by six bundle scripts and six symlink-setup scripts.

## What is in it

**The unification (A–D).** cadgen carries the JS it executes (`_runtime/{node,browser,viewer}`); the viewer backend moves in (`viewer/server_py` → `cadgen/viewer`, mostly R100 renames); skills stop vendoring and their parsers move verbatim into `cadgen/cli/`; the `cadgen` console script arrives; the `earthtojake/cad-viewer` mirror is retired. Phase C alone is −39.8k lines.

**Consolidation that followed.** implicitjs folded into `packages/cadjs` (the theme-system fork reconciled — cadjs's was a strict superset). The MoveIt2 websocket server removed entirely, with the pose picker and the dead code it left; SRDF group states and joint sliders stay. The `api.py` alias layer deleted and `step_artifacts` renamed to `step_topology_artifact`. STEP scene helpers and `import_step` promoted to the cadgen root (18 names).

**The robot validators moved in.** ~4,700 stdlib-only lines out of `skills/{sdf,srdf,urdf}/scripts` into flat format-prefixed cadgen modules, matching how `step`/`dxf`/`implicit` are already laid out — no `robot/` umbrella, because `findings.py` turned out to be format-agnostic. These three `validate` commands were the last place a skill carried real logic. That also deleted a **third URDF reader**: srdf hand-parsed `<robot>` in ~60 lines purely because a skill may not import another skill.

**Shared plumbing under the three XML parsers.** `cadgen/xml_common.py` — five spellings of "path relative to cwd", three namespace strippers, two tag selectors, three duplicate detectors, and 17 statements of inertia eigenvalue math that existed twice. What only *looked* alike stayed split, each difference stated and pinned by a test.

**One front door, not two.** `cadgen step gen` never consulted the warm daemon — only the
skill launchers did — so the console script was an order of magnitude slower than
`scripts/gen` for the same work, silently (0.70s against 0.07s on an already-warm daemon).
Dispatch now owns both pre-import behaviours the launchers had: the daemon handoff and the
`PYTHONHASHSEED` re-run that keeps a content-addressed DXF package byte-deterministic.

**Viewer localhost hardening.** The viewer sits in Jupyter's threat class (`POST /__cad/artifact` executes a generator) with TensorBoard's defenses (loopback only). Added Host-header validation (DNS rebinding) and a required `x-cadgen-viewer` header on every POST — a cross-origin POST is otherwise a no-preflight "simple request" that fires even though the page cannot read the reply. Deliberately **no token auth**: the URL is the filesystem path and must stay pasteable, which is sound once those two gates hold. Plus opt-in `--open`, adversarial path-containment tests against the CVE shapes this tool category keeps shipping, and a running-instance registry (`cadgen viewer list` / `stop`) modelled on `.tensorboard-info` — its `packageDir` field finally makes "which checkout is answering this port" visible.

## What does not change

```
skills/cad/scripts/gen part.step.py   ->  usage: scripts/gen        (unchanged)
cadgen step gen        part.step.py   ->  usage: cadgen step gen    (new front door)
```

Same arguments, same output, same Viewer URL/JSON contract.

**Three deliberate exceptions**, each verified as the only difference:
- Validator `--format json` is compact — moving into cadgen brought them under the repo's payload-budget policy. Same document, fewer bytes.
- A non-tree paired URDF still only *warns*, but SRDF graph cross-checks are now skipped rather than run against a broken tree.
- Every POST needs the header above.

## Defects found and fixed while gating this

Most by the new installed-mode gate — the only check that runs the CLIs the way a user gets them.

| | |
|---|---|
| **Module eviction ate numpy** | `_interpreter_roots()` classified a `.pth`-reached `site-packages` as model code and evicted it. Fatal for a C extension. |
| **Daemon keyed on a moved path** | Socket and staleness token resolved to the wrong directory post-move. |
| **`worker_client` PYTHONPATH** | Pointed at cadgen's package dir, not its parent. Invisible wherever cadgen is installed. |
| **`cad-python.mjs` escaped its worktree** | Bound the dev backend to another branch's cadgen. |
| **Skills could run the wrong cadgen** | Published skills pin `cadgen==X` and nothing checked the installed version. `enforce_requirements_pin` now does, in all 16 shims. |
| **implicit-cad declared no dependency** | No `requirements.txt` at all, though its shims import cadgen — so its own error message pointed at a missing file. `dxf` shipped `scripts/snapshot` without the `[snapshot]` extra, so rendering died on a playwright ImportError. |
| **`/__cad/reveal` answered 405** | "Reveal in Finder" is live in the UI and no backend ever implemented the route. Now implemented. |
| **`os.execv` does not quote on Windows** | develop fixed this in the dxf launchers (#245: `C:\Program Files\...` arrives as two arguments) while this branch was moving that same re-run into cadgen's dispatch — copying the buggy form. Their guard globs `skills/*/scripts/*/__main__.py`, so it structurally could not see the copy. Fixed the same way and their test widened to cover the dispatcher. |
| **Two I introduced** | A latent `NameError` from patching the wrong `build_parser()` call site; an always-on dead pose-picker button. |

Each has a regression test that fails on the old behaviour.

## Policy tests updated, never weakened

The payload-budget test derives coverage from `json.dumps` usage rather than a hand-written list — which caught the validators and `viewer list` as they arrived. New criterion-based guards: every shim that runs installed cadgen checks the pin; every skill importing cadgen declares it, with `[snapshot]` if it renders. `check-builds.sh` sweeps `skills/` for symlinks directly, which matters most here because Codex plugin installs drop them silently.

## Verification

Full suite (928 + 345 JS, ~1,430 Python across 7 suites), docs, `bundle --check` with the viewer required, `check-builds`, wheel contents, version, symlink layout, and `test-installed.sh`.

End-to-end from the **installed wheel in an empty directory**: STEP gen → GLB, export → STL, DXF gen → drawing package via the packaged Node builder, `inspect refs`, snapshot → PNG, the URDF robot with every link mesh, and the Viewer serving a catalog plus building on demand.

**Renders compared against a develop worktree:** all 7 formats 7/7, **47/47 implicits on both**, coverage matching to 4 decimal places. Validator output diffed before/after every refactor against real and malformed fixtures — byte-identical.

Three quirks confirmed **pre-existing on develop**, identical on both, untouched: implicit renders blank under the Vite dev server; the `--all-implicits` sweep exhausts the browser after ~17 models; a symlinked `/tmp` path fails with "not a component-GLB package".

## Your call

- **Build-test rehearsal** — `Release` with `bump=none`, `target_branch=build-test`. Not dispatched; the publish job changed materially here and has never run.
- **The real release** — you pick the bump. Note `VERSION` is still `0.4.9`; the Release workflow owns that bump.
- Archive `earthtojake/cad-viewer` and delete the `CAD_VIEWER_SYNC_TOKEN` secret.


earthtojake:
Superseded by #273.

This branch is in #273 in full, rebased onto develop 0.4.12 as a linear stack, together with the warm-worker pool from `claude/ocp-worker-pool`. Nothing here is dropped.

Two things worth knowing, since they are not visible in a plain rebase:

- Work that lived only in this branch's five develop-merge commits was recovered explicitly — the issue #245 Windows `subprocess` fix in cadgen dispatch, the issue #241 `replace_atomic` retry in the viewer registry, and a test still importing the deleted `server_py`.
- Stacking with the pool exposed a defect neither branch had alone: the pool registered its dispatch under `cadgen.step_artifact`, which develop renamed to `step_artifact_cli` while both branches were out. Fixed in #273 with the allowlist test that pins it.

The pre-rebase tip of this branch is `c3e05186`, and `claude/cadgen-unification` is left untouched.

PR #244, opened 2026-08-14
cadgen: retry SMB GLB rename sharing violations

### LINKED ISSUE #241: 0.4.x: parallel component build fails on SMB/network drives (WinError 32 on the atomic GLB rename)
Building on a Windows network drive (SMB share) fails reliably as soon as the parallel component build kicks in. The atomic rename of a freshly written component GLB hits `WinError 32` ("file is being used by another process"):

```
[scripts/gen] FAILED: RuntimeError: component 2dd5f34fd41b071a build failed: PermissionError: [WinError 32] ...:
'\\\\<nas>\\...\\cad\\__cadgen__\\models\\praxis_tresen.py\\components\\2dd5f34fd41b071a.glb.tmp6704.6704.1786687171725449800.tmp'
-> '\\\\<nas>\\...\\cad\\__cadgen__\\models\\praxis_tresen.py\\components\\2dd5f34fd41b071a.glb.tmp6704'
[scripts/gen]   raised in .../cadgen/_internal/component_package.py:727
```

Note the source path: it's the inner temp file being renamed onto the outer temp name from `_write_component_glb_atomic`, so this is the rename inside the GLB writer, not the outer one.

**A/B test, same project, same machine, cache wiped before each run** (~84 components, so the default worker count is 8 here):

| run | `CADGEN_COMPONENT_WORKERS` | result |
|---|---|---|
| A | unset (8 workers) | fails after 55s, `WinError 32` |
| B | `1` | builds fine in 45s |

Sequential is not just a workaround, it was actually faster in this case. The same project on a local disk (C:) builds fine either way, so it is specific to the network path. I repeated run A three times - it fails every time, and always on the same component (a plain box, 39 KB GLB, ranked 18th by size, referenced by exactly one occurrence, so it isn't a duplicate-cid race or a "large file" thing).

My guess is the SMB redirector's deferred close: the file handle is still held server-side for a moment after Python closed it, and under parallel load that window is long enough for the immediately following rename to lose. A short retry loop around the `os.replace` (a few attempts, backing off ~50ms, only on `PermissionError`/`OSError` with `winerror == 32`) would probably be enough - that's the usual remedy for this on Windows, and it costs nothing when the rename succeeds on the first try.

Happy to test a patch on my setup if that helps.

**Environment:** Windows 10 Pro 19045, Python 3.11, plugin 0.4.8, 20 logical cores. Source tree on a Synology NAS mapped as a drive letter (SMB); the same tree builds fine when copied to a local disk. Not a viewer/locking issue - no CAD viewer server was running during these tests.

NgoQuocViet2001:
Fixes #241.

Windows SMB can briefly retain the just-closed GLB temp file, causing the inner atomic rename to fail with WinError 32. Retry only that sharing violation with bounded 50/100/200 ms backoff; all other replace errors still propagate immediately.

Validation:
- `python -m unittest tests.python.skills.cad.cadgen.test_glb tests.python.skills.cad.cadgen.test_component_package_build tests.python.packages.cadgen.test_component_package` — 29 passed
- `git diff --check`


earthtojake (approved):
Verified locally (10 tests in test_glb pass) and the policy is right: retry only WinError 32, bounded backoff, everything else propagates immediately — and it is a no-op off Windows, since `winerror` does not exist there. Thank you, and thanks for the A/B in #241 with the cache wiped between runs; that is what made the diagnosis unambiguous.

Merging as-is. One gap I will follow up on rather than hold this for: the same `os.replace` appears at six other places in the build's write path on the same share, so the failure can just move —

| | |
|---|---|
| `_internal/glb.py:145` | this PR |
| `_internal/component_package.py:395` | the OUTER component GLB rename, one per component |
| `_internal/component_package.py:769` | `assembly.json` |
| `_internal/drawing_package.py:281` | drawing descriptor |
| `_internal/implicit_package.py:224` | implicit descriptor |
| `step_artifacts.py:443` | STEP artifact |
| `coordination/record.py:103` | the progress record, written many times per build |

Your own traceback in #241 shows the inner and outer renames sitting next to each other, so the outer one is the next most likely to bite. I will extract your retry policy into one helper and use it at all seven, keeping your tests.

PR #253, opened 2026-08-14
cadgen: drop the dead traversal in _scene_has_assembly_structure

warun7:
## Summary

`_scene_has_assembly_structure` in `packages/cadgen/src/cadgen/step_artifact.py` drives part-vs-assembly classification (`infer_entry_kind`) for STEP models that carry no author-supplied `entryKind` — the path the CAD Viewer's on-demand render-package build takes for imported STEP files. The function's loop looked like a full scene-tree walk but was dead code after the first pop: it can never descend past the root.

## The problem

The original implementation:

```python
def _scene_has_assembly_structure(scene: LoadedStepScene) -> bool:
    stack = list(scene.roots)
    if len(stack) > 1:
        return True
    while stack:
        node = stack.pop()
        if node.children:
            return True        # returns immediately
        stack.extend(node.children)  # dead code: only reached when children is falsy
    return False
```

`stack.extend(node.children)` is unreachable with a truthy `children` list because the function already returned `True`, and when `children` is falsy it extends with an empty list — nothing is ever pushed back onto the stack. A reader (or static analysis) expects a deep traversal, but the code only ever inspects the root node. Per CONTRIBUTING's iteration loop, behavior changes should come with measurable regression coverage — here even the intended deep-walk behavior had none.

## Reproduction

This is a code-correctness/packaging issue rather than an observable runtime failure, so reproduction is a static trace of the original function. For any scene with a single root:

1. `stack = [root]`; the root is popped.
2. If the root has children, the function returns `True` — `stack.extend(node.children)` never runs.
3. If the root has no children, the function returns `False` — the `extend` pushes an empty list and the loop exits.

In both branches, nodes below the root are never visited. A scanner can confirm the dead code directly; the new unit tests pin the intended classification table so any future refactor that actually changes the semantics is caught (see "Tests").

## Why the behavior is still correct

The dead loop is misleading, but the classification it produced was right. In a tree, any descendant of the root implies the root itself has children, so "any node at any depth has children" is equivalent to "the root has children". The shallow check and the intended deep walk agree on every possible scene:

- empty roots → part
- single childless root → part
- root with children (any depth below) → assembly
- multiple roots → assembly

## What was fixed

**`packages/cadgen/src/cadgen/step_artifact.py`** — replaced the dead loop with an explicit root-level check and a docstring stating the deliberately shallow scope:

```python
def _scene_has_assembly_structure(scene: LoadedStepScene) -> bool:
    """True if the scene's product hierarchy has child relationships.

    Multiple roots OR any root with children indicates assembly structure.
    The check is deliberately shallow: any descendant implies a child of the
    root, so a root with children already makes the model an assembly.
    """
    if len(scene.roots) > 1:
        return True
    return any(node.children for node in scene.roots)
```

**`tests/python/packages/cadgen/test_assembly_structure_detection.py`** (new) — focused unit tests per CONTRIBUTING ("add focused tests when behavior changes so regressions are measurable"), building `LoadedStepScene`/`OccurrenceNode` directly:

- empty scene → part
- single childless root → part
- root with children → assembly
- root with grandchildren → assembly (deep-structure case)
- multiple roots → assembly

## Verification

- `./.venv/bin/python -m unittest tests.python.packages.cadgen.test_assembly_structure_detection` — 5/5 pass
- Neighboring suites: `test_step_artifact`, `test_step_artifact_skip` — pass (6 tests)
- `scripts/test/test-python.sh` — full green: cadgen package 253, skills (bambu-labs 37, cad 362, dxf 58, gcode 10, sdf 41, srdf 23, urdf 44), moveit2 server 17, viewer backend 174

## Notes

- No `VERSION` bump: per CONTRIBUTING.md, normal development PRs keep `VERSION` untouched (release versions are reserved for release PRs).
- Branched from and targets `develop`, as required for development PRs in CONTRIBUTING.md.


earthtojake:
Reviewed and approved on the substance: the equivalence argument holds — in a tree, any descendant with children implies the root has children, so the shallow check and the intended deep walk agree on every scene — and the five classification tests pin it.

Two things needed before it can merge, both from `develop` moving underneath it:

**1. #257 landed and renamed the module.** `cadgen/step_artifact.py` is now `cadgen/step_artifact_cli.py`. Git's rename detection carries your change to `_scene_has_assembly_structure` across on its own (I merged `develop` into this branch locally to check — the code merges clean, no conflict). The one thing the merge cannot know about is the new test module, which still imports the old name:

```
tests/python/packages/cadgen/test_assembly_structure_detection.py:10: ImportError: cannot import name 'step_artifact' from 'cadgen'
```

Six lines, all in that file:

```diff
-from cadgen import step_artifact  # noqa: E402
+from cadgen import step_artifact_cli  # noqa: E402
```

and the five call sites `step_artifact._scene_has_assembly_structure(...)` → `step_artifact_cli._scene_has_assembly_structure(...)`. With that, the module's 5 tests pass.

**2. The branch needs updating with `develop`** — protection here requires branches be up to date, and several PRs have merged since you opened this.

Tiny nit while you are in there: `test_assembly_structure_detection.py` has no trailing newline.

Ping me once it is pushed and I will run it through.

warun7:
Applied both requested changes:

**1. Module rename**  the new test module now imports the renamed module (6 lines): `from cadgen import step_artifact_cli` plus the five `step_artifact_cli._scene_has_assembly_structure(...)` call sites. Git's rename detection carried the `_scene_has_assembly_structure` change into `step_artifact_cli.py` cleanly on the merge.

**2. Up to date with `develop`**  merged the current tip in (commit `838671f1`); the branch is now up to date and reflects #257's rename plus everything else that landed since opening.

**Nit** trailing newline added to the test file.

Validation: 5/5 tests green on the merged branch (worktree run against the branch's own `packages/cadgen`).

@earthtojake ready for you to run it through.

Issue #246, opened 2026-08-14
dxf: are dimensioned drawings in scope? cut-profile validation blocks them


M-Stege:
Not a bug - the closed-contour rule is documented in `references/generator-templates.md` and my drawing simply predates it. I'd just like to ask whether dimensioned drawings are meant to be in scope.

What I have is a workshop drawing for a cabinetmaker: plan view, two elevations, two sections, a title block and 37 dimensions. Nothing in it is a cut path, so every layer trips `open_cut_profile` and generation stops. That's the layer classifier working as designed - my layers are German (FRONTEN, KORPUS, REFERENZ), and `REFERENZ` doesn't match the `ref` token since matching is on whole tokens. But renaming wouldn't be a language thing either: SECTION, HIDDEN or TITLEBLOCK wouldn't match anything either, so the same wall is there for an English drawing.

I could append tokens (`REFERENZ_REF`) - but the file is already with the workshop and the layer names are what they read in AutoCAD, so I'd rather not reshuffle them for a validator.

The skill description mentions "2D drawing exports of CAD geometry" alongside the cut layouts, which is what made me think drawings might be intended here. If they are, would some way to mark a source as a drawing rather than a cut layout be welcome - a flag in `gen_dxf()`, a module-level constant, anything that skips the cut-profile check while keeping the rest? If drawings are out of scope, that's a fine answer too and I'll keep generating that one file with the older release.

For context, this is the same 0.4.9 setup as #241 and #245.

earthtojake:
Thanks — and you found the real shape of it. This isn't a language thing: `SECTION`, `HIDDEN` and `TITLEBLOCK` match nothing either, so an English drawing hits the same wall. Renaming your layers would have been the wrong fix.

**Landed now** (#250): the classifier knows a drawing's furniture — `section`, `hidden`, `centre`/`center(line)`, `phantom`, `title(block)`, `border`, `frame`, `viewport`, `hatch`, `dim(ension)(s)`, `text`, `label(s)`, `leader`, `axis` — all annotation, none required to close. Whole-token matching is unchanged, so `PREFORM` is still not `ref` and `SECTIONAL` is not `section`. It also fixes an ordering accident: `CUT_SECTION` is a cut path whose name mentions a view, and it was being classified as annotation, skipping the closure check on the layer that most needs it.

[...]

PR #256, opened 2026-08-14
cadgen: annotate or narrow every bare exception handler

warun7:
## Summary
Audits every bare exception handler in cadgen so intentional robustness
is distinguishable from bug-hiding: 39 handlers across 11 files, each
either narrowed to the real failure type or annotated with the existing
`# noqa: BLE001 - reason` convention.

## Changes
- Narrowed where the failure mode is known:
  - drawing_checks.py: SPLINE endpoint reads (AttributeError),
    $INSUNITS parsing (TypeError/ValueError)
  - generation.py: optional build123d/OCP imports (ImportError x2)
- Annotated the intentional catch-alls (OCP C++ exceptions on odd
  shapes, best-effort teardown/cache housekeeping, degrade-to-fallback
  loads, color-type probing, curved-face/edge counting) and added
  rationale comments to the two BaseException run wrappers in
  coordination (a cancelled run is a failed run).

## Not changed
- No behavior changes: narrowed types only drop coverage the bodies
  could not have hit; every annotated handler keeps its existing body.

## Validation
- Zero unannotated `except Exception:` remain under packages/cadgen/src/cadgen
- python -m compileall packages/cadgen/src/cadgen
- scripts/test/test-python.sh (174 tests)

### DIFF
--- a/packages/cadgen/src/cadgen/_internal/component_package.py
+++ b/packages/cadgen/src/cadgen/_internal/component_package.py
@@ -197,7 +197,7 @@ def _bbox_from_shape(shape: Any) -> dict[str, list[float]] | None:
             "min": [float(xmin), float(ymin), float(zmin)],
             "max": [float(xmax), float(ymax), float(zmax)],
         }
-    except Exception:
+    except Exception:  # noqa: BLE001 - OCP bounds reads can raise on odd shapes; a component without bounds is None
         return None
 
 
--- a/packages/cadgen/src/cadgen/_internal/drawing_package.py
+++ b/packages/cadgen/src/cadgen/_internal/drawing_package.py
@@ -186,7 +186,7 @@ def _deterministic_dxf_output(document: object):
             # same document (e.g. a --dxf export) keep real provenance.
             try:
                 previous_created = metadata[ezdxf_document.CREATED_BY_EZDXF]
-            except Exception:
+            except Exception:  # noqa: BLE001 - ezdxf metadata API drift; a missing created-by marker degrades gracefully
                 previous_created = None
             metadata[ezdxf_document.CREATED_BY_EZDXF] = fixed_marker
         yield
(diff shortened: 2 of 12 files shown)


earthtojake:
The 37 annotations and the `ImportError` / `$INSUNITS` narrowings all look right to me — thank you for the rationale on each one, that is exactly the distinction worth making.

One narrowing needs widening before this goes in: `_open_endpoints` in `drawing_checks.py`, narrowed to `AttributeError`. The body reads `control_points` and indexes each point, so a malformed SPLINE raises `TypeError` or `IndexError` there, not `AttributeError` — and those now escape `validate_drawing_document` instead of skipping the entity. Run against this branch versus `develop`:

| `control_points` | develop | this branch |
|---|---|---|
| `[1.0, 2.0]` (scalars, not points) | `None` | `TypeError: 'float' object is not subscriptable` |
| `[(0,), (1,)]` (short tuples) | `None` | `IndexError: tuple index out of range` |
| `[None, None]` | `None` | `TypeError: 'NoneType' object is not subscriptable` |

`ezdxf` yields `Vec3`, so this is not reachable through a well-formed file today. It matters because that handler's job is surviving arbitrary third-party DXFs and `ezdxf` API drift — the case #246 was about — where degrading to "skip this entity" beats failing the whole validation.

One word fixes it, and keeps the narrowing honest about what the body can throw:

```python
except (AttributeError, TypeError, IndexError):  # a malformed SPLINE has no usable endpoints
```

With that changed this is good to merge. Note it will need a rebase first: #250 and #252 have since landed on `develop` and both touch `drawing_checks.py`.

warun7:
Rebased onto current `develop` (includes #250/#252) and applied the requested widening: `except (AttributeError, TypeError, IndexError)` on `_open_endpoints`. Added a regression test covering all three malformed `control_points` shapes from the table (scalars, short tuples, `None` points), plus the still-working usable/closed SPLINE cases — 27/27 drawing-checks tests green.

PR #259, opened 2026-08-14
Add skill-level tests for sendcutsend, step-parts, and implicit-cad

warun7:
## Why

Every other skill (`cad`, `dxf`, `gcode`, `bambu-labs`, `urdf`, `srdf`, `sdf`) has a Python skill-level test suite; these three had none, so nothing in CI pins their SKILL.md frontmatter, references, agents, or script entry points.

## What changed

Six new test files under `tests/python/skills/<skill-dir>/` (19 tests total):

- **`sendcutsend/`** — structure tests (frontmatter, `agents/openai.yaml`, `references/official-sources.md` + `report-template.md`) and a report-template validator (required sections in order: Context → Sources Checked → Geometry Facts → Findings → Verdict; standard findings-table columns)
- **`step-parts/`** — structure tests plus a network-free CLI smoke test (`--help` exits 0, flags documented; bad flag exits non-zero) and a stdlib-only import guard on `download_step_part.py`
- **`implicit-cad/`** — structure tests (frontmatter, agents, script entry points `gen`/`export.mjs`/`snapshot`/`lib/implicit-cad.mjs`, `gen` runnable as a module) and schema-reference tests pinning that the SKILL.md `schema.js` path resolves and carries the `implicit.js/0.1.0` marker

One deviation from the scan plan: test dirs use the skills/ directory names (`step-parts`, `implicit-cad` — hyphens) rather than the planned underscores, because `scripts/test/test-python.sh` keys discovery off `list-skills.sh` basenames; underscored dirs would be silently skipped.

## Verification

- `scripts/test/test-python.sh` full suite green: `sendcutsend` 6, `step-parts` 6, `implicit-cad` 7, and all pre-existing suites unchanged
- No `__init__.py` added, matching the convention of the existing six skill suites

### DIFF
--- a/tests/python/skills/implicit-cad/test_schema_reference.py
+++ b/tests/python/skills/implicit-cad/test_schema_reference.py
@@ -1,0 +1,28 @@
+from __future__ import annotations
+
+import unittest
+
+from tests.python.support.paths import repo_path
+
+SCHEMA_PATH_REFERENCE = "scripts/packages/implicitjs/src/lib/implicitCad/schema.js"
+HELPER_REEXPORT = "scripts/lib/implicit-cad.mjs"
+SCHEMA_MARKER = "implicit.js/0.1.0"
+
+
+class ImplicitCadSchemaReferenceTests(unittest.TestCase):
+    def setUp(self) -> None:
+        self.skill_root = repo_path("skills", "implicit-cad")
+
+    def test_skill_md_names_the_schema_source_of_truth(self) -> None:
+        source = (self.skill_root / "SKILL.md").read_text(encoding="utf-8")
+        self.assertIn(SCHEMA_PATH_REFERENCE, source)
+
+    def test_schema_path_reference_resolves(self) -> None:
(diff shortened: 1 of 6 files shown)


earthtojake:
Looks good — merging this one.

I did not take the green tick at face value on whether these tests can actually fail, so I mutated a fixture: renaming `## Verdict` to `## Renamed Heading` in `skills/sendcutsend/references/report-template.md` produces

```
AssertionError: 'Verdict' not found in ['Renamed Heading'] : section 'Verdict' missing or out of order
```

so the section-order check is load-bearing rather than decorative. Also confirmed all three suites are picked up by `scripts/test/test-python.sh` — `implicit-cad` 7, `sendcutsend` 6, `step-parts` 6 — and every pre-existing suite is unchanged, and that the `step-parts` CLI test really is network-free (`--help` and an invalid flag both exit inside argparse).

Good call on the hyphenated directory names, and on documenting why in the PR body — the runner keys discovery off `list-skills.sh` basenames, so underscored dirs would have been skipped silently.

PR #260, opened 2026-08-14
Add Windows msvcrt fallback to the coordination lock

warun7:
## Why

The generation lock wraps `fcntl.flock` and degrades to a complete no-op where `fcntl` doesn't exist — i.e. Windows. Two concurrent viewer builds there would race the same render package with no coordination at all.

## What changed

`packages/cadgen/src/cadgen/coordination/lock.py` gains a `msvcrt` byte-range backend alongside `fcntl` (both imported behind `ImportError` guards, so the kernel-owned close-releases contract is identical):

- `locking_available()` and every gate now accept either backend; with neither, behavior stays the documented all-degraded no-op
- `probe()`: on Windows takes a 1-byte **exclusive** region lock (`msvcrt` has no shared mode); an **empty sentinel reads as degraded, never held** — lock-past-EOF is an error on Windows, and a crash between `open()` and stamping must not wedge every later build forever
- `_acquire()`: Windows never uses `LK_LOCK` (its ten 1-second retries would raise on a genuinely long wait) — it polls `LK_NBLCK` so the existing deadline and wait-notice semantics work unchanged on both backends
- `_pad_sentinel()`: makes a fresh 0-byte sentinel lockable at all
- Module + `ProbeResult` docstrings updated; the two documented Windows tradeoffs (exclusive probes can false-positive "held"; position-based region locking) are called out

Note: the scan plan described a simple 40-line `lock.py`; the real module is the evolved coordination lock (probe states, bounded waits, re-entrancy, run-id stamping), so the fallback was designed against that, preserving every documented contract.

## Verification

- 9 new tests driving the Windows backend through a faithful fake (`fcntl=None` + fake `msvcrt` keyed by file identity): free/held/empty-sentinel probes, run-id stamping, real waiting acquire, bounded `Contended`, and the fully degraded no-backend state
- 52 coordination tests pass (POSIX path untouched: real 4-process contention, SIGKILLed holder, thread serialization all green)
- `scripts/test/test-python.sh` full suite + `bundle.sh --check` green

### DIFF
--- a/packages/cadgen/src/cadgen/coordination/lock.py
+++ b/packages/cadgen/src/cadgen/coordination/lock.py
@@ -25,6 +25,16 @@ Readers probe with ``LOCK_SH``, writers take ``LOCK_EX``. That asymmetry matters
 conflicts per open file description, not per process, so two concurrent ``LOCK_EX`` probes
 of an UNHELD sentinel conflict with each other and one of them wrongly reports a build in
 flight. Measured at ~6% false positives with four threads before this was fixed.
+
+On Windows there is no ``fcntl``; :mod:`msvcrt` provides byte-range locks (``locking``)
+instead. Kernel-ownership on close behaves the same, but the model differs in two ways
+that matter: ``msvcrt.locking`` locks a byte region at the CURRENT file position (not the
+whole descriptor), and it has no shared mode -- every operation, probes included, takes an
+exclusive region lock. A sentinel therefore must hold a byte before it can be locked at
+all, and two concurrent Windows probes of the same sentinel can false-positive "held"
+(the very race the POSIX shared-probe asymmetry exists to avoid). Locking past EOF is an
+error on Windows, so an EMPTY sentinel is reported as degraded rather than held -- a
+crash between ``open()`` and stamping must not wedge every later build forever.
 """
 
 from __future__ import annotations
@@ -38,10 +48,19 @@ import uuid
 from pathlib import Path
 from typing import Callable, Iterator, NamedTuple
 
-try:  # POSIX only; on a platform without fcntl every operation degrades to a no-op.
+try:  # POSIX only; on a platform without fcntl every operation uses the windows backend
+    # or degrades to a no-op.
     import fcntl
 except ImportError:  # pragma: no cover - not reachable on darwin/linux CI
     fcntl = None  # type: ignore[assignment]
+
+try:  # Windows only; msvcrt.locking is a byte-range lock, not a whole-descriptor lock.
+    import msvcrt
+except ImportError:  # pragma: no cover - not reachable on darwin/linux CI
+    msvcrt = None  # type: ignore[assignment]
+
+
+_BUSY_ERRNOS = (errno.EWOULDBLOCK, errno.EAGAIN, errno.EACCES)
 
 
 _HELD = threading.local()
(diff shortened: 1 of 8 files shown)


earthtojake:
The backend design is right and the research behind it shows: no shared mode, region-at-current-position, lock-past-EOF being an error, and `LK_LOCK`'s ten one-second retries are each handled, and polling `LK_NBLCK` keeps the deadline and wait-notice semantics identical across backends. The fake is honest too — keyed by `(st_dev, st_ino)`, so cross-handle contention is real rather than stubbed. 24 coordination tests green here.

Two things before merge, one of them a real leak on POSIX rather than Windows.

**1. `_pad_sentinel`'s byte is visible to `read_run_id`, and the pad runs on POSIX.**

The docstring says it "writes nothing visible to `read_run_id` (which strips whitespace)", but `"\x00"` is not whitespace, so `str.strip()` keeps it:

```
after _pad_sentinel, read_run_id -> '\x00'
```

And POSIX reaches it: `_acquire` calls `_pad_sentinel` in the polling branch — i.e. any caller passing `deadline_ms` or `on_wait`, which is the viewer's normal path — before the lock is held. So on a filesystem that refuses advisory locks, which is the NFS/SMB case #241 and #244 were about, `flock` raises `ENOLCK`, `exclusive()` degrades exactly as documented, and nothing ever stamps the sentinel:

```
degraded as designed, recorded run id: None
sentinel bytes: b'\x00' -> read_run_id '\x00'
```

That `'\x00'` then flows into `Snapshot.run_id` and into `progress_for_run(...)` as an attribution key. Two small changes fix it:

- pad with `b" "` instead of `b"\x00"` — a space *is* stripped, which makes the docstring's claim true;
- call `_pad_sentinel` only on the `msvcrt` branch. POSIX `flock` is whole-descriptor and never needed a byte to lock, so this also keeps "the POSIX path is untouched" literally true.

**2. `_BUSY_ERRNOS` widens the POSIX busy set.**

`EACCES` is the Windows busy errno; folding it into the shared tuple means a POSIX `flock` raising `EACCES` is now read as "held" in `probe()` and as "keep polling" in `_acquire` instead of raising. Per-backend tuples (`EWOULDBLOCK`/`EAGAIN` for `fcntl`, plus `EACCES` for `msvcrt`) keep each contract exactly as it was.

Minor observation, no action needed: `_FakeMsvcrt` does not model lock-past-EOF as an error, so the empty-sentinel test passes because production checks the size first, not because the fake enforces the rule. Worth knowing if that branch is ever refactored.

One sequencing note: this branch contains #259's six test files, so #259 goes in first. You will also need to update this branch against `develop` — protection here requires branches be up to date, and several PRs have landed since.

warun7:
Both required fixes are in, committed as `f86ea00b` on `feature/windows-file-locking`:

1. **Sentinel pad** — `_pad_sentinel` now writes `b" "` instead of `b"\x00"` (invisible to `read_run_id`'s whitespace strip, so no `'\x00'` run id leaks into status records), and it is now called only on the msvcrt branch: the POSIX polling path (`deadline_ms`/`on_wait`) no longer touches the sentinel file at all (`_acquire`, lock.py:288).

2. **Per-backend busy errnos** — `_BUSY_ERRNOS` split into `_FCNTL_BUSY_ERRNOS` (`EWOULDBLOCK`, `EAGAIN`) and `_MSVCRT_BUSY_ERRNOS` (+ `EACCES`), used by both `probe()` branches and the `_acquire` poll loop. POSIX `flock` raising `EACCES` now raises instead of polling forever.

Verified: all 52 tests in `test_coordination.py`, `test_coordination_lock.py`, and `test_coordination_is_stdlib_only.py` pass with `CADGEN_STRICT_LOCKS=1`.

@earthtojake — ready for another look when you get a chance. Note the branch is still behind `develop`; I'll sync it once #259 lands.

PR #265, opened 2026-08-15
feat(dfam-check): skill for per-process DfAM printability measurement

knewnothing-git:
Adds `dfam-check`, a skill that reports printability facts for a mesh before slicing.

**What it measures:** mesh integrity, down-facing overhang maps, ray-cast wall thickness, support volume estimate, and a 6-orientation ranking. Two subcommands, `measure` and `orientations`.

**Rules:** 26 numeric limits — 6 categories across 5 processes (FDM/FFF, SLS, SLA/DLP, PBF-LB metal, MJF), with 4 `n/a` cells where powder support makes the rule meaningless. They live in `references/process-limits.md` as a table rather than in code, so a user with a machine- or material-specific datasheet can override one cell and have the workflow cite that source instead of a hardcoded default.

**Design:** `dfam_tool.py` measures and never emits pass/fail. Verdicts belong to the SKILL.md workflow, which keeps thresholds inspectable rather than buried in the tool.

**Standards:** findings cite ISO/ASTM 52910 and 52911-1 clause numbers alongside independently measured values. The documents themselves aren't vendored — they're copyrighted by ASTM, so only clause references are used.

**Local only:** no API key, no network call. Adds `rtree` to `requirements-dev.txt`; `trimesh` and `numpy` are already present.

**Tests:** parts are built with build123d so the expected answer comes from the geometry rather than a recorded run — a 2 mm wall measures 2 mm because it was built that way, and a 30 deg underside is below a 45 deg limit by definition. Covers overhang flagging either side of the limit, plate-face exclusion, wall thickness magnitude and ordering, the orientation ranking finding the self-supporting flip, and both CLI subcommands. 9 tests, ~3s, no new dependency since build123d is already a dev requirement.

Sits ahead of `$gcode`: check printability, then slice.

Happy to add a README badge if you want one — left it alone rather than presume.


warun7:
Minor observations from a review pass. No verdict implied, leaving that to the maintainer.

1. `support_area_pct` uses total surface area as its denominator. It works as a signal, but a projected-area denominator would be more comparable across orientations. The `method` field already documents it as coarse, so this is a documentation nit at most.

2. `_wall_facts` does `rng.choice` with `p=mesh.area_faces / mesh.area_faces.sum()`. A mesh with zero total face area would crash that call. It is unreachable through a valid file load in practice, but a guard or a comment would make the assumption explicit.

3. The skill description advertises .stl, .obj, .ply, and .3mf inputs. The tests only exercise STL. The claimed formats load through `trimesh.load`, but coverage would confirm it.

Also worth noting: the tests run in about 0.1 s locally, not the roughly 3 s the PR body mentions. That is a pleasant surprise, not a problem.

knewnothing-git:
> Minor observations from a review pass. No verdict implied, leaving that to the maintainer.
> 
> 1. `support_area_pct` uses total surface area as its denominator. It works as a signal, but a projected-area denominator would be more comparable across orientations. The `method` field already documents it as coarse, so this is a documentation nit at most.
_1 — the denominator. Agreed it's a documentation nit, and I've documented rather than changed it. support_area_pct now ships with a pct_denominator field and a comment explaining the tradeoff: total surface area is rotation-invariant, so the values stay comparable across candidates, which is what the ranking needs. Projected area would be more physically meaningful per-orientation but isn't comparable between them. The guidance is to rank on support_area_mm2 and read pct as a signal. Happy to switch if you'd rather have the physical quantity._
> 2. `_wall_facts` does `rng.choice` with `p=mesh.area_faces / mesh.area_faces.sum()`. A mesh with zero total face area would crash that call. It is unreachable through a valid file load in practice, but a guard or a comment would make the assumption explicit.
_2 — the divide by zero. Real bug, fixed. _wall_facts now checks the total face area before building the sampling weights, and a mesh with no positive area returns that fact rather than crashing inside rng.choice. You're right that a valid file load won't reach it, but the assumption was invisible, and a tool that measures shouldn't fall over on a mesh it can't measure. Added a test with three collinear vertices._
> 3. The skill description advertises .stl, .obj, .ply, and .3mf inputs. The tests only exercise STL. The claimed formats load through `trimesh.load`, but coverage would confirm it.
_3 — the format gap. This was the one I should have caught. SKILL.md advertises .stl, .obj, .ply, .3mf and only STL was exercised. Now all four are, round-tripped through trimesh and checked for volume and body count. A format that's claimed but unreadable fails at the user instead of in CI, which is the worse place for it._
> 
> Also worth noting: the tests run in about 0.1 s locally, not the roughly 3 s the PR body mentions. That is a pleasant surprise, not a problem.
_On timing — you're right that my ~3s figure was loose. It moves around depending on whether build123d is already imported, so I'd rather not put a number on it than put a wrong one._

earthtojake:
Thanks @knewnothing-git for building this. It fills a real gap in the lineup (nothing between `$cad` and `$gcode` checked printability) and the shape is right for this repo: the tool measures, the SKILL.md decides, the limits sit in a table a datasheet can override, and the tests derive expected answers from constructed geometry. The fixes for warun7's findings landed in code, and they hold up.

I ran the tool beyond the unit tests: a clean venv built from the skill's own `requirements.txt` (trimesh 5.0.0), plus the repo's mesh fixtures. That covered multi-body assemblies (9, 11, and 42 bodies), a 127k-triangle staircase, meters-scale robot-arm STLs, a gear with 60 faces deleted to fake a broken export, a planar mesh, and all four advertised formats. Two of those runs found blockers.

## Blockers

**1. `requirements.txt` is missing `scipy`, `networkx`, and `lxml`.**
With only `trimesh numpy rtree` installed, the first `measure` of any mesh dies with a raw `ModuleNotFoundError`: `mesh.body_count` (dfam_tool.py:44) imports scipy for connected components. `.3mf`, which SKILL.md advertises, fails to load without networkx, then lxml. CI can't catch this because `requirements-dev.txt` installs all three for other suites. With the three added, every advertised format loads and measures. Fix: add them to `skills/dfam-check/requirements.txt` and mirror them under the `# - skills/dfam-check/requirements.txt` comment in `requirements-dev.txt`.

**2. Multi-body meshes report clearance gaps as wall thickness.**
`models/mesh/3mf/basic_shape_mating_test_fixture.3mf` (11 bodies) measures `min_mm: 0.091`, `p05_mm: 0.4`. Split per body, the thinnest wall in that file is 0.745 mm. Rays cast from one body cross the mating clearance and hit the neighbour, so the tool records the gap as a wall. SKILL.md tells the agent to treat `p05_mm` below the limit as a violation, which turns every tight-clearance assembly into a false wall-thickness failure. Most of our fixtures are multi-body. Fix: sample per body via `mesh.split(only_watertight=False)` and aggregate, or reject hits landing on triangles from a different body.

## Should fix

**3. Planar meshes crash `measure` with a qhull traceback and no JSON.**
A single-triangle STL loads, has positive area, passes the degenerate-area guard, then dies in `_support_volume_facts` at `m.convex_hull.volume` (`QH6214 ... not enough points`). `measure` assembles the whole report before printing, so the user gets none of the facts that did compute. A per-family try/except that degrades to an `"error"` field in that section would cover this and anything else one family throws.

**4. Meters-scale meshes read as a perfect print.**
`models/robots/so101/meshes/sts3215_03a_v1.stl` uses meter units (bbox 0.05 × 0.02 × 0.04). The 0.1 mm on-plate tolerance exceeds the part height, so all 3,425 down-facing faces count as on-plate: overhangs 0.0%, support 0.0, and `volume_mm3` rounds to 0.0. The SKILL.md unit check catches this if the agent reads the bbox, but the numbers themselves look like a pass. A measured hint (flag when the bbox diagonal is under ~1) would give the workflow a hook instead of a judgment call.

**5. The skill is missing from the README table and the docs page.**
The other eleven skills appear in both the README skills table and the array in `docs/src/app/page.tsx`. One row and one entry; fine to fold into this PR.

**6. Three limits-table rows have no measured counterpart.**
Min hole diameter, min positive feature, max unsupported bridge. The trapped-powder gap gets an explicit "report ❓" in SKILL.md; these three don't, and the eyeballing ban only names walls, overhangs, and support. A note in `process-limits.md` marking which rows the tool measures today would close that loophole.

**7. `Closes #188` points at nothing.** Issue 188 404s in this repo. Drop or correct it before merge.

Everything else held up: 11 tests in 0.08 s, the 8 mm mounting plate measures an 8.0 mm median wall, STL and 3MF of the planetary assembly agree on volume and body count to the decimal, 127k triangles measure in 1.5 s, the broken-export gear degrades to `watertight: false` with the hull fallback, and the orientation ranking finds the staircase's on-side orientation at a third of the support area. With the two blockers fixed this is a strong addition. Thanks again for the care in it.

knewnothing-git:
Thanks for the thorough pass — the clean-venv run especially. All seven addressed, force-pushed just now (rebased onto current develop, so the diff is 10 files, +883, no deletions).

Blockers

1 — missing deps. Reproduced your clean-venv failure exactly: mesh.body_count → connected_components → scipy, then .3mf → networkx → lxml. Added all three to the skill's requirements.txt. Only scipy is mirrored into requirements-dev.txt; networkx and lxml are already under the repo-tests block, so mirroring them would duplicate — same way trimesh is handled. Confirmed all four advertised formats load and measure in a venv built from the skill's own requirements.

2 — clearance read as wall. _wall_facts now splits with mesh.split(only_watertight=False) and measures each body against itself, then pools, with a per_body breakdown and a measured_per_body flag. Sample budget is shared by surface area with a floor of 64 so small bodies still get sampled. On a socket/peg mating fixture it now reports the 1.7 mm socket wall and 6.0 mm peg separately instead of the 0.3 mm gap. Split cost is ~150 ms on an 81k-face mesh (7493 → 7643 ms) — the bulk of that time is trimesh's pure-Python ray fallback, which is also why you saw 127k in 1.5s and I see 81k in 7.5s; an environment with the embree backend collapses it.

Should-fix

3 — planar crash. Each fact family now runs through a _safe wrapper and degrades to an error field. The planar mesh returns valid JSON with support_volume carrying the qhull error while mesh facts and overhangs still compute.

4 — meters-scale. New scale block on both subcommands: bbox_diagonal_mm and units_suspect (true under 1 mm). SKILL.md now branches on the measured field instead of eyeballing the bbox.

5 — README + docs. Added to both the README table and the page.tsx array, positioned before G-code.

6 — unmeasured rows. process-limits.md now names which rows have a measured counterpart (walls, self-supporting angle) and which don't (hole diameter, positive feature, unsupported bridge), with instruction to report the latter as not-checked rather than infer them.

7. Dropped Closes #188 — you're right, it 404s here.

One implementation note: pooling per-body thickness meant passing raw numpy arrays between functions. Those use underscore-prefixed keys that are stripped before output, and there's a test asserting they never leak and the payload stays JSON-serialisable.

To your rule-count question from earlier: 26 numeric limits — 6 categories × 5 processes, with 4 n/a where powder support makes the rule moot. Tests are 17, built on parts where the answer is known by construction.

earthtojake:
Both blockers clear. I built a venv from only `skills/dfam-check/requirements.txt` and a full `measure` runs on STL and 3MF with no import error, and `basic_shape_mating_test_fixture.3mf` now reports 11 bodies with `measured_per_body: true`. The old pooled 0.091 mm artifact is gone. Planar meshes return valid JSON, the so101 STL trips `units_suspect`, the skill is listed in the README and the docs page, and the three unmeasured limits are documented as not-findings rather than inferred. 17 tests pass.

One thing to fix first. `per_body` entries carry `min_mm` and no `p05_mm`, and that puts two rules in SKILL.md against each other:

- Line 69 says to judge on `p05_mm`, because `min_mm` alone can be a sampling outlier.
- Lines 71-73 say to attribute an assembly violation to its own body, and that a pooled figure is not a finding.

On an assembly the agent can follow one or the other. Per-body gives it the number line 69 distrusts; pooled gives it the number line 73 rejects.

It bites on your own fixture. Measuring the bodies directly:

```
body 0: min=0.436  p05=7.0
body 8: min=0.081  p05=3.408
```

Body 0 reads as a violation against FDM's 0.8 mm limit on a wall that is 7 mm thick. That is the false fail the per-body split was added to remove, arriving from a stray ray inside one body rather than a ray crossing between two.

`_wall_facts_single` already computes the value, so the builder just needs to pass it through:

```python
per_body.append({
    "body": i,
    "min_mm": facts.get("min_mm"),
    "p05_mm": facts.get("p05_mm"),
    "median_mm": facts.get("median_mm"),
    "samples_valid": facts.get("samples_valid", 0),
})
```

Also `Closes #188` in the PR body points at an issue that does not exist.

With those two, this is good to merge.

knewnothing-git:
Good catch — and yeah, that's a real bug, not a nitpick. _wall_facts_single was computing p05_mm the whole time, I just never wired it into per_body. So the two rules were fighting over a field that only half-existed.

Fixed with your one-liner — per_body now carries p05_mm right next to min_mm. Ran it back through your fixture logic to be sure: on a 2-body mating pair, both bodies report p05_mm cleanly through the CLI, so the agent can judge FDM's 0.8mm limit per body without ever touching the pooled figure. Added a test that just checks p05_mm shows up on every per_body entry, so this can't quietly regress if I touch that function again.

Also dropped Closes #188 from the description.

18 tests passing now. Appreciate you actually running this in a clean venv instead of taking the SKILL.md at its word — that's twice now you've caught something I'd have shipped otherwise.

earthtojake:
Merged. Thanks @knewnothing-git for building this, and for rerunning it in a clean venv each round.

One follow-up when you get a chance. The `if __name__ == "__main__": unittest.main()` block sits at `tests/python/skills/dfam-check/test_dfam_tool.py:215`, with `MultiBodyWallTest`, `ResilienceTest` and `ScaleHintTest` defined below it. Running the file on its own executes 11 of 18 tests and prints OK, and the 7 it skips include the `per_body` `p05_mm` coverage we just added. CI calls `python -m unittest <paths>`, which imports the module, so all 18 ran here and nothing slipped through on this PR. Moving the block to the end of the file keeps a direct run honest.

I also dropped `Closes #188` from the description on merge.

Issue #269, opened 2026-08-15
Windows: DXF generation always fails — the sentinel can't be read while its own lock is held


M-Stege:
0.4.11 fixed the file-URL problem from #266, so the Node builder starts now. It gets one step further and then dies every time:

```
Error: render package written without its generation lock: ...\__cadgen__\models\praxis_tresen_werkstatt.py
(--run-id d128d312572c4d7ca0b1de7d3d806956 does not match ...\models\.praxis_tresen_werkstatt.py.generation.lock)
    at assertWriteLock (dxf-artifact.mjs:2:122683)
```

The confusing part is that the sentinel on disk contains exactly the run id the message says doesn't match. I checked it right after the failure — same 32 hex characters, byte for byte.

The reason is that nobody can read that file while the build holds it. `lock.py` takes `msvcrt.locking(handle.fileno(), msvcrt.LK_NBLCK, 1)` on byte 0 of the sentinel and holds it for the whole run. Windows byte-range locks are mandatory, not advisory like `flock`, so any other process reading the file gets EBUSY. `assertWriteLock` reads it with `fs.readFileSync`, the `catch` turns that into an empty string, and the comparison against `--run-id` then fails. On POSIX the same code works because `flock` doesn't block readers at all — which is probably why this survived a fix that was verified from a POSIX box.

Minimal repro, no cadgen involved:

```python
import msvcrt, os, subprocess
p = os.path.abspath("sentinel")
open(p, "wb").write(b"d2ca1a2fcb44472a8ab64a946c2b3247")
js = ("try{console.log('OK', require('fs').readFileSync(process.argv[1]).toString())}"
      "catch(e){console.log('FAILED', e.code)}")
h = open(p, "a+b"); h.seek(0)
msvcrt.locking(h.fileno(), msvcrt.LK_NBLCK, 1)
print(subprocess.run(["node", "-e", js, p], capture_output=True, text=True).stdout)  # FAILED EBUSY
msvcrt.locking(h.fileno(), msvcrt.LK_UNLCK, 1)
print(subprocess.run(["node", "-e", js, p], capture_output=True, text=True).stdout)  # OK d2ca...
```

Moving the locked byte out of the run-id region does not help, I tried that first: `readFileSync` reads the whole file and hits the locked byte wherever it sits. A positional `fs.readSync(fd, buf, 0, 32, 0)` of just the run-id region does succeed, so that would be a JS-side fix.

[...]

Issue #266, opened 2026-08-15
0.4.10: Node builder fails on Windows — file:// + pathname2url makes an invalid five-slash URL


M-Stege:
Building a DXF on Windows with 0.4.10 dies in the Node child before it writes anything:

```
[scripts/gen] FAILED: NodeBuilderError: Node builder failed with exit code 1: dxf-artifact.mjs
TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute
    at getPathFromURLWin32 (node:internal/url:1490:11)
    at fileURLToPath (node:internal/url:1614:35)
    at finalizeResolution (node:internal/modules/esm/resolve:243:12)
  code: 'ERR_INVALID_FILE_URL_PATH',
  input: 'file://///C:/Users/.../cadgen/_internal/node_resolve_register.mjs',
```

Five slashes. It's the `--import` bootstrap URL from `node_runtime.py:197`:

```python
return "file://" + pathname2url(str(path))
```

On Windows `pathname2url` already returns a leading `///`, so the prefix doubles up. On POSIX it returns `/usr/...` and the same line is correct — which is presumably why it went unnoticed.

```python
>>> pathname2url(r'C:\Users\...\node_resolve_register.mjs')
'///C:/Users/.../node_resolve_register.mjs'
>>> "file://" + pathname2url(p)     # what's built now
'file://///C:/Users/...'
>>> Path(p).as_uri()                # what Node accepts
'file:///C:/Users/...'
```

`Path.as_uri()` handles both platforms and also percent-encodes, which matters here since the path can contain spaces. `"file:" + pathname2url(...)` would work too.

No workaround from the outside as far as I can see — the URL is built internally and `--import` is always passed.

The same line is in every vendored copy of `node_runtime.py` (the cad skill's included), so it should hit any Node builder invocation on Windows; a STEP build happens not to reach it here, so I only see it via `skills/dxf/scripts/gen`.

Unrelated but from the same run: #252 works exactly as intended on my drawing — `drawing_document: treated as a dimensioned drawing (37 dimension/leader entities)`, no layer renamed. This is the only thing now standing between that fix and a regenerated DXF.

**Environment:** Windows 10 Pro 19045, Python 3.11, Node v22, plugin 0.4.10. Same failure on a local disk and on a mapped network drive, so it isn't UNC-related.

earthtojake:
Fixed in **0.4.11**, published just now.

[...]

PR #272, opened 2026-08-15
ci: run the test suite on Windows, and cover the bugs that got there first

earthtojake:
Four of the last five user-reported bugs were Windows-only — #260, #266, #267, #269 — and every one of them passed CI. `test.yml` has only ever run `ubuntu-latest`, and the entire Windows-specific surface in shipped code is three files:

| file | what only runs on Windows | shipped a bug |
|---|---|---|
| `coordination/lock.py` | the `msvcrt` byte-range backend | #260, #269 |
| `_internal/atomic_replace.py` | the `ERROR_SHARING_VIOLATION` retry | — |
| `_internal/node_runtime.py` | the `file://` bootstrap URL | #266, #267 |

None of it has ever executed in CI.

## The coverage was mostly already written

`test_coordination.py`'s `test_reports_writing_while_a_peer_holds_the_lock` spawns a real holder subprocess and asserts the parent can read the run id out of the sentinel — **exactly** the cross-process read #269's mandatory lock blocks. It is not a Windows test; it passes on Linux today and would have failed on Windows with the bug present. What was missing was a runner, not a test.

## A separate job, never a matrix

A matrix on `test` renames its check to `Test (ubuntu-latest)`, and `Test` is a required status check on `develop`. It would stop being reported and every pull request would wait forever for a check that no longer exists.

The job runs the **tests** only, not the hygiene steps the Linux job wraps them in. Bundle freshness, the symlink layout and the published-tree checks are properties of the repository, not of an operating system; bundling is deterministic, so re-checking it on a second OS costs minutes and buys nothing. The platform risk is in file I/O — locks, paths, subprocesses, file URLs — which is what the suites exercise.

`core.symlinks` is set before checkout: Git for Windows otherwise materializes symlinks as text files holding the target path, and `develop` is a symlink layout by design. Without it the runtime paths under `skills/` are plain files and the suite tests nothing real.

## Regressions now asserted against the real backend

Everything in `WindowsLockBackendTests` drives a fake, because `msvcrt` is not importable on the box that has always run these — and a fake cannot reproduce either bug Windows users actually hit. The new `RealBackendRegressionTests` asserts the invariants directly. On POSIX they pass trivially; on a Windows runner they have teeth:

- **#260 — a backend that silently is not there.** `exclusive()` yields None rather than failing, which is right for a filesystem that cannot lock and wrong as a description of a whole platform. Any OS the tests run on is an OS someone builds on, so a missing backend is a bug there, not a degradation.
- **#269 — a lock that makes its own file unreadable.** Asserted FROM ANOTHER PROCESS, the only place mandatory and advisory locking differ: the holder can always read its own locked region.

**#266/#267** already had this shape and needed nothing: `test_node_resolve_bootstrap.py` pins both platform flavours from one box through `PureWindowsPath` and `nturl2path`, and scans every vendored copy for the concatenation pattern. On a Windows runner its first two tests become real rather than simulated.

## Two tests had to be fixed before they could run there

- The unwritable-directory test now **skips** on Windows. `chmod` only toggles the read-only bit there and does not stop a directory being written, so it would have passed *without ever reaching the degradation it claims to cover*. A vacuous pass is worse than a failure: it reports coverage it is not providing.
- The "currency check runs under the lock" test probed with a bare `import fcntl` — an ImportError on Windows, where the assertion is worth more, not less. It now goes through `lock.probe()`, which asks the same question through whichever backend is compiled in.

## Deliberately not a required check yet

It should run green for a while first. The open question is whether `cadquery-ocp` installs cleanly on `windows-latest` — it publishes wheels, so it should, but this run is how we find out rather than assert it.

Promoting it later is a one-line branch-protection change, and worth doing deliberately: under `develop`'s `strict` up-to-date rule, a second required check makes every merge in a stack as slow as the slower runner.

## Verification

cadgen **280** (+2), full Python suite green on Linux. The Windows job's result is the point of opening this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
           node scripts/release/sync-version.mjs --check
 
   test:
-    name: Test
+    name: Test (Linux)
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,3 +60,38 @@ jobs:
 
       - name: Run code tests
         run: scripts/test/test.sh
+
+  # A SEPARATE JOB, never a matrix on `test` above. A matrix renames that job's check to
+  # "Test (ubuntu-latest)", and `Test` is a required status check on develop -- it would stop
+  # being reported and every pull request would wait forever for a check that no longer exists.
+  #
+  # Runs the TESTS only, not the hygiene steps the Linux job wraps them in. Bundle freshness,
+  # the symlink layout and the published-tree checks are properties of the repository, not of
+  # an operating system; bundling is deterministic, so re-checking it here would buy nothing
+  # and cost minutes. The platform risk is in file I/O -- locks, paths, subprocesses, file
+  # URLs -- which is what the test suites exercise.
+  #
+  # Why this job exists at all: four of the last five user-reported bugs were Windows-only
+  # (#260, #266, #267, #269), and every one of them passed CI. The coverage was mostly already
+  # written -- test_coordination.py's holder test asserts exactly the cross-process sentinel
+  # read that #269 broke -- so what was missing was a runner, not a test.
+  test-windows:
+    name: Test (Windows)
+    runs-on: windows-latest
+
+    steps:
+      # Before checkout: Git for Windows materializes symlinks as text files containing the
+      # target path unless this is set, and `develop` is a symlink layout by design. Without
+      # it the runtime paths under skills/ are plain files and the suite tests nothing real.
+      - name: Enable Git symlinks
+        run: git config --global core.symlinks true
+
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up dependencies
+        uses: ./.github/actions/setup-deps
+
+      - name: Run code tests
+        run: scripts/test/test.sh
+        shell: bash
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,6 +92,20 @@ jobs:
       - name: Set up dependencies
         uses: ./.github/actions/setup-deps
 
-      - name: Run code tests
-        run: scripts/test/test.sh
+      # Three steps rather than test.sh, each running even if an earlier one failed, and the
+      # Python runner in --keep-going mode. While Windows is being brought up the failures are
+      # independent POSIX-isms in unrelated suites, and stopping at the first one turns a list
+      # into one ~10 minute round trip per entry. The Linux job keeps the fail-fast default.
+      - name: Run JS tests
+        run: scripts/test/test-js.sh
+        shell: bash
+
+      - name: Run Python tests
+        if: ${{ !cancelled() }}
+        run: scripts/test/test-python.sh --keep-going
+        shell: bash
+
+      - name: Run global policy tests
+        if: ${{ !cancelled() }}
+        run: scripts/test/test-global.sh
         shell: bash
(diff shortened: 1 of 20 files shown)


earthtojake:
@M-Stege — #274 is fixed here, and so is the reason you had to find it yourself.

**On the measurement.** You were right that the old window looked correct and wasn't. `RETRY_DELAYS_SECONDS` is now `(0.05, 0.1, 0.2, 0.4)`, five attempts over 750 ms, exactly as you proposed. What I've written into the docstring is your argument rather than just the number, because that is the part a future reader needs: the budget is spent per rename, but the build only succeeds if *every* rename wins, and a large assembly draws from that distribution a hundred-odd times. 350 ms genuinely was generous per rename — your median is 31 ms — and still lost roughly one build in five at 168 draws. Your figures (126 of 168 blocked, median 31, p90 118, max 389) are now in `test_atomic_replace.py`, so the number can't shrink back without meeting them.

Your negative result saved me from a wrong turn, so it's recorded too: a synthetic write-then-rename loop produces zero sharing violations, and the measurement only exists because you took it from the real build path.

And no apology needed for opening it separately. A measurement posted to a closed thread is a measurement nobody reads — you filed it correctly.

**On why these keep reaching you.** #260, #266, #267, #269 and now #274 were all Windows-only, and every one of them passed CI, because `test.yml` has only ever run `ubuntu-latest`. This PR adds a `Test (Windows)` job, and it is now green across the full suite.

Most of the coverage already existed. `test_coordination.py`'s holder test asserts precisely the cross-process sentinel read that #269 broke — it passes on Linux and fails on Windows with that bug present. What was missing was a runner, not a test. The job found four more things on its first runs, which are also fixed here:

- `_resolve_cli_output_path` rejected any path containing a backslash, so **no native Windows path** could be passed to `--output` or to a `SOURCE=OUTPUT` target. If you ever tried `cad gen part.step.py=D:\models\part.step` and gave up, that's why.
- `step_targets` guards against a target outside the command cwd, and the guard never fired on Windows: it keyed on `is_absolute()`, which is False for a rooted path with no drive, so the bogus-relative-path failure it exists to prevent was still live there.
- `urdf.classify_mesh_uri` called a rooted mesh path *relative* on Windows, resolving it against an unrelated base and silently producing a path nobody wrote.
- `scripts/gen --write` emitted `parts\second.step` into a logical `SOURCE=OUTPUT` pair, where the form is POSIX.

One thing the job surfaced that is **not** fixed here, in case it matters to you: the cadgen daemon speaks over an AF_UNIX socket in `/tmp`, and neither half exists on Windows. Its tests skip there now rather than pretending otherwise. Say if you need the daemon and I'll open that separately.

Nothing here is a substitute for you running it on real hardware against a real NAS — CI has no SMB share, so #274's failure mode remains invisible to it. But the class of bug that has been costing you a release cycle each time should mostly stop arriving now.

M-Stege:
No need to open the daemon one on my account. I checked before answering: nothing in my build chain touches it, and the viewer server doesn't either — the word doesn't appear anywhere in `server_py`. Skipping those tests on Windows is the right call for me.

The CI finding is the more interesting half of your comment from where I sit. "What was missing was a runner, not a test" explains the whole run of Windows-only issues better than any of the individual bugs did, and the backslash rejection in `_resolve_cli_output_path` would have found me eventually — my chain only escapes it because it passes bare filenames in the working directory.

On #274: agreed that your CI can't see it, so I'll run the real gate here once it's in a release. Dropping `CADGEN_COMPONENT_WORKERS=1` and doing a few full parallel builds against the NAS is the only thing that answers it, and I'll report the result either way, including if 750 ms still isn't enough.

Issue #274, opened 2026-08-15
SMB: the 350 ms rename retry window covers the median but not the tail, so large assemblies still fail


M-Stege:
Apologies for raising this twice. I already posted these numbers as a comment on #241 after you had closed it, so you may well have read them once already. Closed threads are easy to miss, which is the only reason I am opening a separate issue rather than leaving the measurement buried there. If you would rather keep the discussion in the closed thread, say so and I will not push it again.

Follow-up to #241, which you closed as fixed in 0.4.10. The retry absolutely helped: before it, the parallel build failed every single time and always on the same component. After it, 1 of 3 runs completed and the other two died on different components. But it still fails here more often than not, and I think the numbers explain why.

I measured the actual wait rather than guessing. I temporarily widened `RETRY_DELAYS_SECONDS` to fine steps up to 10s, logged how long each rename really had to wait, ran two full builds (84 components, 8 workers) on the SMB share, then reverted the patch:

```
blocked renames: 126 of 168 total
min 15 ms | median 31 ms | p90 118 ms | max 389 ms
```

The median is comfortably inside the current window. The tail is not: 389 ms against a 350 ms budget.

The docstring reasons about this per rename, and per rename it is right — "a rename that cannot win in that window is not a deferred close" is a sound rule for one operation. What the current value misses is that the budget is spent per rename but the build only succeeds if *every* rename wins. With 168 renames, a per-rename failure probability of even 1% leaves roughly a 1-in-5 chance of a clean build. The window has to cover the tail of the distribution, not its middle, because a large assembly draws from that distribution a hundred-odd times.

Concretely, one more step would do it:

```python
RETRY_DELAYS_SECONDS = (0.05, 0.1, 0.2, 0.4)   # five attempts over 750 ms
```

That doesn't slow down the common case at all, since a median wait of 31 ms wins on the first or second retry, and 126 of my 168 blocked renames never went past 200 ms. It only changes what happens for the handful that hit the tail. The bound stays bounded, which I take to be the actual point of the docstring — a build that hangs retrying really is worse than one that says what happened.

Worth saying explicitly: this is not urgent for me. `CADGEN_COMPONENT_WORKERS=1` sidesteps it completely, and sequentially the build is actually *faster* here (45 s against 55 s to the abort), so I have a working setup either way. It only bites parallel builds on a network share; local disks never block at all in my runs.

[...]

PR #283, opened 2026-08-16
cadgen: rewrite under a fresh temp name when the rename ladder is exhausted

NgoQuocViet2001:
Picks up @M-Stege's first suggestion in #274, and your reply that if 750 ms still is not enough then the fix is a different strategy at the rename site rather than a bigger constant.

Their 0.4.13 numbers are the case for it: 413 of 2067 renames blocked, median 31 ms, p90 78 ms, p99 265 ms, max 797 ms — one rename beat the window, and one is all it takes. The tail is long and thin, so any finite window eventually loses to it.

## What changed

`write_bytes_atomic` now writes the payload twice at most. If `replace_atomic` exhausts its ladder on `WinError 32`, the payload is written again under a **fresh** temp name and the ladder runs once more. The violation is pinned to the handle the server still holds on the temp file just closed, so a file it has never seen carries no deferred close and the second ladder starts clean.

Still bounded — one extra write, not a longer window — which is the principle the window itself was chosen under.

`RETRY_DELAYS_SECONDS` is untouched, so the measured-tail assertions still hold. `replace_atomic` is unchanged: callers that hand it an existing file own no payload and cannot rewrite it, so they keep the plain ladder.

I have no SMB share to verify the real-world effect, so this is the mechanism argument plus unit coverage — the same gap your comment names about CI. Worth @M-Stege re-running their parallel gate against it if they are willing.

## Test plan

- Ran: `PYTHONPATH=".:packages/cadgen/src" python -m unittest tests.python.packages.cadgen.test_atomic_replace` (11 passed, 3 new)
- Checked: with `WRITE_ATTEMPTS = 1` the new exhaustion case fails, so it is pinned to the behaviour and not the wiring
- Ran: the whole `tests/python/packages/cadgen` set — same failures as a clean `develop` on this checkout (no `OCP` installed here; `test_concurrent_generation` is flaky either way)

New cases: the retry after exhaustion uses a temp name the first ladder never used; two ladders then the original error propagates and no temp file is left behind; any error other than `WinError 32` still fails on the first write.

### DIFF
--- a/packages/cadgen/src/cadgen/_internal/atomic_replace.py
+++ b/packages/cadgen/src/cadgen/_internal/atomic_replace.py
@@ -37,17 +37,30 @@ p99 265 ms -- and max 797 ms. Exactly one rename in 2067 beat the window, and on
 takes. The tail is long and thin, so any finite window eventually loses to it and chasing it
 trades a bounded wait for a slow creep.
 
-So ``write_bytes_atomic`` attacks the cause instead of the budget: when the ladder is exhausted
-it writes the payload once more under a FRESH temp name and runs the ladder again. The violation
-belongs to the handle the server still holds on the temp file just closed; a file it has never
-seen carries no deferred close. One extra write in the rare tail case, still strictly bounded --
-which is the same principle the window was chosen under. Callers that own an existing file, and
-so cannot rewrite it, keep the plain ``replace_atomic`` ladder.
+So the ladder is not the last word: when it is exhausted, ``replace_atomic`` copies the temp file
+sideways under a name the server has never seen and runs the ladder once more on the copy. The
+violation belongs to the handle the server still holds on the ORIGINAL temp file; the deferred
+close pins that handle, reads do not conflict with it, and the copy carries no handle of its own.
+One extra copy in the rare tail case, still strictly bounded -- the same principle the window was
+chosen under.
+
+It lives here rather than in ``write_bytes_atomic`` because most of these renames have no payload
+to rewrite. A measured gate run on the reporter's NAS put every remaining failure at the component
+GLB rename, which streams its bytes to disk and hands this helper a finished file; five other call
+sites do the same, and a branch in flight adds a seventh. Hardening the writer would have left all
+of them out, which is the rule the docstring above already states.
+
+What this cannot do: ``os.replace`` is also refused when something holds the DESTINATION open --
+a reader with the artifact mapped, or a virus scanner sweeping it. No source-side trick reaches
+that case, and the second ladder is simply a bounded futile wait when it happens. The reported and
+measured failure is the source handle; this covers that one.
 """
 
 from __future__ import annotations
 
+import contextlib
 import os
+import shutil
 import time
 from pathlib import Path
 
@@ -56,22 +69,62 @@ from pathlib import Path
 WINDOWS_SHARING_VIOLATION = 32
 RETRY_DELAYS_SECONDS = (0.05, 0.1, 0.2, 0.4)
 
-# How many times ``write_bytes_atomic`` writes the payload before giving up: the first write
-# plus one rewrite under a fresh temp name. See the module docstring -- this is the bounded
-# answer to the long tail, in place of a longer window.
-WRITE_ATTEMPTS = 2
+
+def temp_suffix() -> str:
+    """A collision-free suffix for a sibling temp file, ending in ``.tmp``.
+
+    ``os.urandom`` rather than a clock: two writes inside one clock tick would otherwise collide,
+    and the rescue below depends on the new name being genuinely new rather than on the timer's
+    resolution. The ``.tmp`` ending matters too -- ``.gitignore`` carries ``*.tmp``, so a temp
+    file leaked in the deep tail stays invisible to ``git add -A``.
+    """
+    return f".{os.getpid()}.{os.urandom(4).hex()}.tmp"
 
 
-def replace_atomic(temp_path: Path | str, target_path: Path | str) -> None:
-    """``os.replace`` with a bounded retry for the SMB sharing violation."""
+def _run_ladder(temp_path: Path | str, target_path: Path | str) -> OSError | None:
+    """Run the bounded retry, returning the sharing violation it could not beat.
+
+    Any other error propagates at once: a denial (5) or a missing directory is a real error and
+    must surface immediately, not 750 ms later.
+    """
     for delay in (*RETRY_DELAYS_SECONDS, None):
         try:
             os.replace(temp_path, target_path)
-            return
+            return None
         except OSError as error:
-            if getattr(error, "winerror", None) != WINDOWS_SHARING_VIOLATION or delay is None:
+            if getattr(error, "winerror", None) != WINDOWS_SHARING_VIOLATION:
                 raise
+            if delay is None:
+                return error
             time.sleep(delay)
+    return None
+
+
+def replace_atomic(temp_path: Path | str, target_path: Path | str) -> None:
+    """``os.replace`` with a bounded retry, then one retry from a copy the server has not seen."""
+    blocked = _run_ladder(temp_path, target_path)
+    if blocked is None:
+        return
+
+    source = Path(temp_path)
+    fresh = source.with_name(f"{source.name}{temp_suffix()}")
+    try:
+        shutil.copyfile(source, fresh)
+    except OSError:
+        # The source is pinned for reading too, so there is nothing left to try. Report the
+        # rename failure, which is the one the caller was waiting on.
+        raise blocked from None
+
+    # Best effort: the handle that blocked the rename usually blocks this as well.
+    with contextlib.suppress(OSError):
+        source.unlink(missing_ok=True)
+
+    try:
+        if _run_ladder(fresh, target_path) is not None:
+            raise blocked
+    finally:
+        with contextlib.suppress(OSError):
+            fresh.unlink(missing_ok=True)
 
 
 def write_bytes_atomic(target_path: Path, payload: bytes) -> None:
@@ -79,23 +132,15 @@ def write_bytes_atomic(target_path: Path, payload: bytes) -> None:
 
     Same directory on purpose: a rename across filesystems is not atomic, and a temp dir on
     another volume would silently turn this into a copy.
-
-    If the rename ladder is exhausted by a sharing violation, the payload is written once more
-    under a fresh temp name and the ladder runs again. The violation is pinned to the handle the
-    server still holds on *that* temp file, so a file it has never seen starts clean rather than
-    re-colliding with the same stale handle.
     """
     resolved = Path(target_path)
     resolved.parent.mkdir(parents=True, exist_ok=True)
-    for attempt in range(WRITE_ATTEMPTS):
-        temp_path = resolved.with_name(f"{resolved.name}.{os.getpid()}.{time.time_ns()}.tmp")
-        try:
-            temp_path.write_bytes(payload)
-            replace_atomic(temp_path, resolved)
-            return
-        except OSError as error:
-            last_attempt = attempt == WRITE_ATTEMPTS - 1
-            if last_attempt or getattr(error, "winerror", None) != WINDOWS_SHARING_VIOLATION:
-                raise
-        finally:
+    temp_path = resolved.with_name(f"{resolved.name}{temp_suffix()}")
+    try:
+        temp_path.write_bytes(payload)
+        replace_atomic(temp_path, resolved)
+    finally:
+        # Suppressed: on Windows the handle that blocks a rename blocks the delete too, and
+        # letting that escape here would mask the rename failure the caller needs to see.
+        with contextlib.suppress(OSError):
             temp_path.unlink(missing_ok=True)
--- a/packages/cadgen/src/cadgen/_internal/atomic_replace.py
+++ b/packages/cadgen/src/cadgen/_internal/atomic_replace.py
@@ -30,6 +30,19 @@ times. At 168 renames even a 1% per-rename loss rate leaves roughly a 1-in-5 cha
 build, which is what the reporter saw: 1 of 3 runs completing, each failure on a different
 component. Widening costs the common case nothing, because a 31 ms median still wins on the
 first or second retry.
+
+The 750 ms window is not the end of it, and a bigger constant is not the answer. Remeasured on
+0.4.13 across eight instrumented builds: 413 of 2067 renames blocked, median 31 ms, p90 78 ms,
+p99 265 ms -- and max 797 ms. Exactly one rename in 2067 beat the window, and one is all it
+takes. The tail is long and thin, so any finite window eventually loses to it and chasing it
+trades a bounded wait for a slow creep.
+
+So ``write_bytes_atomic`` attacks the cause instead of the budget: when the ladder is exhausted
+it writes the payload once more under a FRESH temp name and runs the ladder again. The violation
+belongs to the handle the server still holds on the temp file just closed; a file it has never
+seen carries no deferred close. One extra write in the rare tail case, still strictly bounded --
+which is the same principle the window was chosen under. Callers that own an existing file, and
+so cannot rewrite it, keep the plain ``replace_atomic`` ladder.
 """
 
 from __future__ import annotations
@@ -42,6 +55,11 @@ from pathlib import Path
 # ago and the server has not caught up.
 WINDOWS_SHARING_VIOLATION = 32
 RETRY_DELAYS_SECONDS = (0.05, 0.1, 0.2, 0.4)
+
+# How many times ``write_bytes_atomic`` writes the payload before giving up: the first write
+# plus one rewrite under a fresh temp name. See the module docstring -- this is the bounded
+# answer to the long tail, in place of a longer window.
+WRITE_ATTEMPTS = 2
 
 
 def replace_atomic(temp_path: Path | str, target_path: Path | str) -> None:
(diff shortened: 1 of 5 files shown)


M-Stege:
Ran the parallel gate against this branch on the real NAS, as asked. Method: applied your `atomic_replace.py` onto the vendored cadgen in 0.4.14, added one temporary marker line in the rewrite branch so I could tell whether the fallback actually fired (rare-event runs that all pass prove nothing on their own), ran 13 full parallel builds (84 components, 8 workers), reverted everything afterwards.

**The mechanism works.** The rewrite branch fired **5 times** across the 13 runs, each time on a component GLB write, and every rescued rename went on to complete — five builds that would have died under plain 0.4.14 didn't. Your handle-pinning argument is confirmed by behaviour, not just by reasoning: the fresh temp name won immediately every time.

**But 3 of 13 builds still failed — all at the same uncovered call site.** Every remaining failure is the *outer* component rename, `component_package.py:728`, renaming `<hash>.glb.tmp<pid>` → `<hash>.glb`:

```
PermissionError: [WinError 32] ... 'components\7cb2891999babc93.glb.tmp46400' -> 'components\7cb2891999babc93.glb'
  raised in …/cadgen/_internal/component_package.py:728
```

That call goes through `replace_atomic` directly, so it keeps the plain ladder. The split is perfectly clean in my logs: all 5 rescues at the inner `write_bytes_atomic` rename, all 3 failures at the outer direct `replace_atomic` — which is exactly the "one frame away" pair the module docstring already describes.

(Side note on rates: the NAS was having a much worse day than during my 0.4.13 measurement — 8 tail events in ~13 builds against 1 in 8 the day before. Bad luck for my pass rate, good luck for the experiment: it exercised both paths properly.)

On your rationale for leaving `replace_atomic` untouched — "callers that hand it an existing file own no payload and cannot rewrite it": true for the payload, but the *file* is still rewritable. At this call site the source is a complete, just-written temp file; on ladder exhaustion it can be **copied to a fresh temp name** and the ladder run again on the copy. The deferred close pins the old handle, reads don't conflict with it, and the copy carries no deferred close — same trick, one level up. That would close the remaining 3-of-13 gap with the same bounded shape: one extra copy in the tail case.

So from where I sit: the PR does what it says and I'd take it as-is — it turned five would-be failures into passes here. It just doesn't close #274 by itself while `component_package.py:728` keeps the bare ladder. Happy to re-run the gate against an updated branch.

Environment: branch `fix-smb-rewrite-on-exhaustion` @ 382c984 applied over 0.4.14, Windows 10 Pro 19045, Python 3.11, Synology SMB share on a mapped drive.

earthtojake:
Thanks for this — the diagnosis is the right one. Attacking the cause rather than widening the window is the correct instinct, and the remeasured tail (p99 265 ms, max 797 ms) is a much better argument than picking a bigger constant. Two things need fixing before it can land, and the first is what your red Windows build is already telling you.

### 1. The fresh temp name isn't fresh

`Test (Windows)` fails on this PR's own test:

```
AssertionError: 'artifact.glb.9116.1786905406838327800.tmp' unexpectedly found in
{'artifact.glb.9116.1786905406838327800.tmp'}
: the retry after exhaustion must use a temp file the server has never seen
```

The set has exactly one element, and the second attempt's name equals it. The name is built as `{name}.{pid}.{time_ns()}.tmp`, and since the test mocks out `time.sleep`, both attempts land inside a single clock tick — so `time.time_ns()` returns the same value and the "fresh" file is the same file.

Reproduced locally by pinning the clock:

```
os.replace calls    = 6
distinct temp names = 1
```

That defeats the premise. The retry hands the server back the exact file the violation was pinned to, so the second ladder re-collides with the same stale handle by construction.

In production the ladder's real sleeps make a collision unlikely — but the whole argument for this change is "a file it has never seen," and that shouldn't rest on clock resolution. Folding the attempt index (or a few bytes of `os.urandom`) into the name makes the guarantee structural and fixes the build at the same time.

### 2. The cleanup can pre-empt the retry

`finally: temp_path.unlink(missing_ok=True)` now sits between the two attempts. On Windows, deleting a file whose handle is still open raises `ERROR_SHARING_VIOLATION` — the same condition that blocked the rename. When that fires it propagates out of the `finally` and the second attempt never runs, so the fix is a no-op in exactly the scenario it was written for.

Modelling that (rename refused *and* delete refused, which is one server-side cause):

```
os.replace calls    = 5      <- one ladder, then out
distinct temp names = 1
```

The guard is one line:

```python
finally:
    with contextlib.suppress(OSError):
        temp_path.unlink(missing_ok=True)
```

Verified: the retry becomes reachable (10 calls / 2 distinct names) and all 11 tests in `test_atomic_replace.py` still pass. Worth noting the unlink was already inside a `finally` before this PR, so the masking isn't new — it just matters much more now that the retry has to get past it.

### One limitation worth recording

`os.replace` can be refused because of a handle on the **destination** as well as the source — a reader holding the artifact open, or an AV scanner. A fresh temp name can't help there; it just buys a second futile ladder. That doesn't argue against the change (it's still bounded, and the source-handle case is the reported one), but the module docstring currently presents the source handle as the whole story, and a sentence acknowledging the other case would keep the reasoning honest for whoever reads it next.

Happy to re-review once the temp name is unambiguous and the unlink is guarded.

earthtojake:
Following up on my review above with a change to the ask, after modeling both findings and sitting with @M-Stege's gate numbers.

@NgoQuocViet2001, the two fixes stand, and I'd rather you fold them into a reshape than patch them in place: move the whole fallback into `replace_atomic`, as @M-Stege proposed, and let `write_bytes_atomic` revert to its four-line form. The copy variant is the same mechanism you built. It covers the sites your version cannot reach.

The case for the move comes from this thread's own data:

- All three remaining gate failures sit at `_write_component_glb_atomic`, which streams the GLB to disk through `build_component_glb_from_shape`. That writer holds no payload in memory, so a payload rewrite can never cover it, and five other call sites hand `replace_atomic` a temp file the same way.
- The module docstring called this outcome in advance: "there are seven of these renames in a build's write path. Hardening one moves the failure to the next." The test class guarding the module is named for the same rule, one policy rather than one per writer.
- A branch in flight adds an eighth site (`viewer/registry.py` on `release/0.5.0`). It goes through `replace_atomic`, so a helper-level fallback covers it with zero coordination. A writer-level one leaves it out.

Modeling finding 2 also turned up a sharper argument for the unlink guard than the one I gave. Your rescue can only fire after the between-attempts unlink succeeds, which means the server had already dropped its hold during the ladder. Nothing in the five rescued renames shows the fresh name doing work a same-name retry could not have done. The case where the fresh name earns its keep, a handle held past 750 ms, is the case the unguarded unlink aborts. Until the guard lands, the mechanism has never run in the scenario it was designed for.

The shape I have in mind (sketch, names negotiable):

```python
def replace_atomic(temp_path, target_path):
    error = _run_ladder(temp_path, target_path)      # today's 750 ms ladder
    if error is None:
        return
    # The ladder lost to a handle the server holds on temp_path. A file it
    # has never seen starts clean: copy sideways, run one more ladder.
    fresh = temp_path.with_name(
        f"{temp_path.name}.{os.getpid()}.{os.urandom(4).hex()}.retry.tmp")
    shutil.copyfile(temp_path, fresh)
    with contextlib.suppress(OSError):
        temp_path.unlink(missing_ok=True)            # best effort; may be pinned
    try:
        if _run_ladder(fresh, target_path) is not None:
            raise error                              # the original failure, as today
    finally:
        with contextlib.suppress(OSError):
            fresh.unlink(missing_ok=True)
```

Details worth carrying over from the review:

- Build the fresh name from the pid plus `os.urandom`, no clock. That makes freshness structural and turns your red Windows test green for the right reason.
- End the name in `.tmp`. Line 36 of `.gitignore` is `*.tmp`, so a copy leaked in the deep tail stays invisible to `git add -A`. Related wart while you are in there: the component writer's `.tmp<pid>` suffix escapes that pattern (`git check-ignore` confirms `foo.glb.tmp123` is eligible for tracking), so a leaked temp at that site can reach a commit today. That suffix deserves the same rename.
- Guard the callers' own `finally: unlink` blocks too. They can mask the original error on the failure path with or without this change.
- The "no temp file left behind" test expectation needs a knowing edit: in the deep tail a temp can now survive. Today that same case leaves the file and fails the build, so both halves of the trade improve.

On blast radius: off Windows the first `OSError` raises at once because `winerror` does not exist there, and the happy path runs no new code on any platform. The visible change confines itself to Windows renames that already fail today, which now report after about 1.5 s instead of 750 ms and hold one transient duplicate temp during the rescue.

Two notes belong in the docstring rather than the code. A handle on the destination side, a reader or a scanner holding the artifact open, stays out of reach for any source-side trick, and the docstring should say so. For the content-addressed `<cid>.glb` files, "destination already exists" could count as success, since the name pins the bytes by construction; that one is a separate PR.

@M-Stege, the copy idea is yours, and your NAS remains the only SMB coverage this project has. Once the branch is reshaped, one more gate run would give us the verification CI cannot. Thank you for both instrumented runs; the 3-of-13 split settled the scope question.

NgoQuocViet2001:
Reshaped as you asked — the rescue now lives in `replace_atomic` and `write_bytes_atomic` is back to its short form. Both findings reproduced here first, so the fixes are pinned to the behaviour rather than to the wiring.

**Finding 1 — the fresh name wasn't fresh.** Reproduced with the clock pinned:

```
before:  os.replace calls = 10,  distinct temp names = 1
after:   os.replace calls = 10,  distinct temp names = 2
```

Names now come from pid + `os.urandom(4)` through a shared `temp_suffix()`, so freshness is structural. `test_temp_suffix_is_unique_without_a_clock` pins 200 draws inside one frozen tick.

**Finding 2 — the cleanup pre-empted the retry.** Modelled as rename refused *and* delete refused:

```
before:  os.replace calls = 5,   distinct temp names = 1   <- one ladder, then out
after:   os.replace calls = 10,  distinct temp names = 2
```

Your sharper argument for this one is right, and it changes what the earlier gate run proved: the five rescues @M-Stege measured could only fire after the unlink succeeded, so none of them exercised a handle held past 750 ms. The mechanism had never run in its own scenario. `test_the_copy_survives_a_pinned_source` covers that now.

**Coverage.** The outer site all three gate failures hit is rescued:

```
components/7cb2.glb.tmp -> components/7cb2.glb   rescued: True, distinct names = 2
```

**Also in.** `.tmp<pid>` at four sites (`component_package` ×2, `step_artifacts`, `coordination/record`) now goes through `temp_suffix()`, so `git check-ignore` accepts them — it rejected `x.glb.tmp123` before. The `finally: unlink` blocks in `step_artifacts` and `component_package` are guarded too; `coordination/record` already was.

**Test-expectation edit, knowingly.** `test_the_temp_file_does_not_survive_a_failure` stays for the ordinary failure. Alongside it, `test_a_pinned_temp_file_may_survive_rather_than_mask_the_failure` records the trade in its name: a genuinely pinned temp can now outlive the call, and the caller gets the rename error instead of a cleanup error.

The docstring now says a handle on the destination — a reader or a scanner holding the artifact open — is out of reach for any source-side trick, and that the second ladder is a bounded futile wait in that case.

Left out on purpose: treating "destination already exists" as success for `<cid>.glb`, since you called that a separate PR.

## Test plan

- Ran: `python -m unittest tests.python.packages.cadgen.test_atomic_replace` — 14 passed (6 new)
- Ran: the whole `tests/python/packages/cadgen` set — 47 failures/errors, identical to a clean `develop` on this checkout, no new names
- Checked: both findings before/after, and the `component_package` rename, with the numbers above
- Checked: `git check-ignore` accepts the new suffix and rejected the old one

@M-Stege — the copy is your idea and the branch is updated if you're willing to run the gate once more. The `component_package.py` rename is covered this time, which is where your 3 of 13 landed. Worth keeping your marker-line trick: with the unlink guarded, a rescue can now fire in the case that previously aborted, so the counts should not be directly comparable to the last run.

M-Stege:
Re-ran the gate against the reshaped branch (b90931e applied over 0.4.14's vendored cadgen, all four source files, same marker instrumentation in the rescue branch of `replace_atomic`, everything reverted afterwards).

**13 of 13 parallel builds passed. Zero `WinError 32` anywhere in the logs.**

The rescue fired **3 times**, and the split is the one that matters:

```
components/71eda2465f89b460.glb                      <- outer rename, the site that killed 3/13 last round
components/60ca2e43d6065b57.glb                      <- outer rename
components/3e80ee59d11cc842.glb.39756.b46a4e46.tmp   <- inner write_bytes_atomic rename
```

Two of the three rescues sit exactly at the outer component rename that produced every failure in my previous run — the site the payload-rewrite version could not reach. Each rescued rename completed and its build went green. No temp files were left behind in the components directory afterwards (84 GLBs, zero `*.tmp*` stragglers), so the deep-tail leak case didn't occur here.

One honest caveat on what this run does and does not prove, given the earlier finding about the unlink: I can't tell from the marker alone whether these three rescues involved a handle held past 750 ms or a handle that had already dropped by copy time. What I can say is behavioural and it's the number that counts for #274: same NAS, same model, same 8 workers — previous branch 10/13 with all failures at the uncovered site, this branch 13/13 with the rescue observably firing at that exact site.

For rate context across the three days of this thread: plain 0.4.10 ≈ 1/3 builds passing, 0.4.13 ≈ 7/8, previous branch 10/13, this branch 13/13 with rescues confirmed at both call-site levels. From where I sit this closes #274 on my hardware.

Environment: branch @ b90931e over 0.4.14, Windows 10 Pro 19045, Python 3.11, Synology SMB share on a mapped drive.

earthtojake:
Looks like some tests are still failing @NgoQuocViet2001 !

NgoQuocViet2001:
Sorry — that's mine, and it's a real one. Fixed in f3dc314.

One failure on each runner, same test: `test_components_are_clean_and_byte_deterministic`. Two builds of the same geometry stopped producing the same bytes.

```
b'…5335.e66703eb.step.glb","runsView":"faceRuns",…'
     != b'…5335.bb6eca69.step.glb","runsView":"faceRuns",…'
```

Those eight hex characters are `os.urandom(4).hex()` from `temp_suffix()`, in the payload. The path is `build_component_glb_from_shape`:

```python
placeholder = out_glb.with_suffix(".step")
scene = build_build123d_step_scene(_unlocated_shape(shape), placeholder)
```

`out_glb` there is the sibling temp file, and the placeholder's name comes back out through `step_scene_selectors.py:993` as `faceProxy.source`. `.tmp<pid>` was a *single* suffix, so `with_suffix()` replaced it whole and the staging name never survived. `temp_suffix()` adds two components and only `.tmp` is the last one:

```
develop      abc.glb.tmp12345            -> .abc.glb.step.glb
this branch  abc.glb.12345.e66703eb.tmp  -> .abc.glb.12345.e66703eb.step.glb
```

So the determinism this test guards was resting on `with_suffix()` eating exactly one component — and, before that, on both builds in the test sharing a pid. Two builds in separate processes would already have disagreed on develop if the suffix had been one component longer.

The fix takes the destination for the placeholder rather than the file being written:

```python
identity = identity_path or out_glb
placeholder = identity.with_name(f"{identity.name}.step")
```

Appended, not substituted, and deliberately: `<cid>.step` reads better, but the placeholder has to stay exactly what the old temp name reduced to or every component GLB changes and the content-addressed caches keyed on them go stale. That is not this PR's business. Verified the embedded name is now both invariant and unchanged from develop:

```
after fix, distinct embedded names across 3 (pid,random) draws: 1
matches develop exactly: True
```

I checked the other three sites the suffix rename touched. Only this one derives content from its output path — `_write_topology_sidecar` and the descriptor write pass the real `step_path` separately, and `coordination/record` writes bytes it already holds.

I can't run this particular test locally (no OCP in this checkout), so CI is the confirmation. The existing test is the right guard and it did its job; I didn't add another one for it.

## Test plan

- Ran: `test_atomic_replace` + `test_coordination` — 42 passed
- Checked: the placeholder derivation across pid/random draws, against develop's, with the numbers above
- Checked: `_write_topology_sidecar`, the descriptor write and `coordination/record` for the same output-path-to-payload path — none of them have it

@M-Stege — no need to spend another gate run on this one; it's a naming bug in the component writer, not the rescue. The rescue path is byte-for-byte what you measured at b90931e.

M-Stege:
Re-ran the gate against f3dc314, same setup as before: four source files over 0.4.14's vendored cadgen, one marker in the rescue branch of `replace_atomic`, everything reverted afterwards and checked byte-identical against the backups. 0.4.15 landed in between, but all four files are byte-identical between 0.4.14 and 0.4.15, so this is exactly your delta and the numbers stay comparable to the earlier rounds.

**13 of 13 parallel builds passed. Zero `WinError 32`, zero temp stragglers.**

The rescue fired **5 times** across 4 of the 13 runs, at both levels again:

```
run 01  967c5e203978480e.glb                      <- outer rename (component_package.py:398)
run 01  3d7aae9a5dffb3c6.glb.46784.223fd595.tmp   <- inner write_bytes_atomic (glb.py:144)
run 03  dba4927b54da86cb.glb.4808.1bcf6770.tmp    <- inner
run 06  b7ede65a85f23cc8.glb.2152.da147ffa.tmp    <- inner
run 09  1d807db64f0f1772.glb.8696.090a6a24.tmp    <- inner
```

Every rescued rename completed and its build went green.

**On the determinism fix specifically**, since that is the part CI can check but not under these conditions: after each green build I hashed all 84 component GLBs (name + content) and compared across runs. **One single hash across all 13 builds.** That includes run 01, where the outer component rename came through the rescue copy — i.e. the file that landed in `components/` was written under a name the build had not planned for, and its bytes still match the twelve builds that never went near the rescue. `identity_path` holds exactly where the old code would have leaked, because the rescue is the one path that hands the writer a genuinely new name.

Honest caveat on the strength of that: only one of the five rescues was at the outer site, so the byte-level evidence at that site rests on a single build. The other twelve hashes agreeing is a strong baseline, but I would not claim to have hammered the outer path. Rescue firings are rare events and I can't provoke them on demand.

Rate context across this thread, same NAS and same model each time: plain 0.4.10 ≈ 1 in 3 builds passing, 0.4.13 ≈ 7/8, first shape of this branch 10/13 with all failures at the uncovered outer site, reshaped branch 13/13, and now 13/13 again with determinism confirmed on top. The NAS was in a middling mood today, 5 tail events against 3 in the previous run and 8 the day before, so the pass rate isn't flattered by a quiet share.

Timing for the record: parallel builds averaged 49.3 s (47.6 to 57.2). The sequential reference build after restoring unpatched 0.4.14 took 45.6 s and my model's own checks pass unchanged (0 collisions, all volume and dimension checks OK). Sequential staying marginally faster is a property of this share, not of your PR.

Environment: branch @ f3dc314 over 0.4.14, Windows 10 Pro 19045, Python 3.11, Synology SMB share on a mapped drive.

earthtojake (approved):
Approving. Both review findings are fixed and verified independently:

- the temp name is now os.urandom-based, not clock-based (5000/5000 distinct draws, no clock in the returned name)
- both cleanup unlinks are suppressed, so a refused delete no longer pre-empts the rescue (verified: rescue runs, 2 distinct names, payload lands)
- bounded in the unwinnable destination-pinned case: exactly 2 ladders, surfaces WinError 32, no temp left behind

The reshape into replace_atomic covers all seven rename sites rather than just the byte writer, and the follow-up correctly caught that the two-component suffix broke with_suffix('.step'). I verified the placeholder fix preserves content addressing: the same model builds to 9 identical component hashes on develop and on this branch, so no cached component is invalidated.

@M-Stege's gate run on f3dc314 is the confirmation CI cannot give: 13/13 builds green on the NAS that reported #274, rescue observably firing at both call sites including the outer one that caused every failure last round, and 84 component GLBs hashing identically across all 13 builds. I checked his baseline claim too — all four files are byte-identical between 0.4.14 and 0.4.15, so his numbers apply to exactly this head.

Thanks to both of you; the instrumented runs are what settled the scope question.

NgoQuocViet2001:
Thanks both — and @M-Stege especially, that's three instrumented gate runs on your hardware for this one change. The 3-of-13 split you measured is what settled the scope question, and without the marker trick neither of us would have known whether the rescue was firing or just not being needed.

For anyone reading this later, the sequence that mattered: the payload-rewrite version couldn't reach `component_package.py:728` because that writer streams to disk and holds nothing to rewrite; moving the rescue into `replace_atomic` covered it; and the unguarded `finally: unlink` meant the mechanism had never actually run in the scenario it was written for until that guard landed.

The determinism failure at the end was mine — `temp_suffix()` added two dot-components where the old `.tmp<pid>` was one, and `with_suffix(".step")` only replaces the last, so the pid and random draw ended up inside the component GLB as `faceProxy.source`. Worth remembering that the component builder derives an embedded name from the path it is handed; it makes any future change to temp naming a payload change too.

PR #292, opened 2026-08-18
fix: wheel zoom jumps from 100% to 10%/800% in a few notches

jarvis959:
## What's wrong

Wheel zoom is unusable. Two or three wheel notches at 100% land you at 10% zoomed out or pin the view at 800%, and `zoomToCursor` makes it feel like the camera is being flung. Scroll-zoom should take many notches to cross that range, not a flick of the wheel.

## Root cause

OrbitControls scales camera distance by `0.95^(zoomSpeed * delta/100)` per wheel event. This viewer handed it `zoomSpeed = 10` for the accelerated wheel path (`ACCELERATED_WHEEL_ZOOM_SPEED`) and `4.5` for the main path (`DEFAULT_ZOOM_SPEED`). Both are orders of magnitude off the upstream default of `1.0`, so one notch moves the camera by a factor far beyond the viewable range.

## The fix

`viewer/src/client/components/CadViewer.js`, constants only:

- `ACCELERATED_WHEEL_ZOOM_SPEED`: 10 → 1.0
- `DEFAULT_ZOOM_SPEED`: 4.5 → 1.0
- `COARSE_POINTER_ZOOM_SPEED`: 1.6 → 1.0
- `TRACKPAD_PINCH_ZOOM_SPEED`: 14 → 2.0 (trackpads send many small deltas, so it stays a little ahead of a plain wheel)
- `COARSE_POINTER_PINCH_ZOOM_SPEED`: 2.4 → 1.6

## Verification

Headless Chromium + Playwright, one wheel notch = delta 120, model = a 150×150×50 mm part (the DGX Spark stand):

- Before: 3 notches out → 10% (model becomes a dot); 5 notches in → 800% (useless perspective).
- After: 6 notches out → ~69%, whole model stays visible; 6 notches back in → returns to ~100%. No blanking at any step.

Numbers came from the same scripted input both before and after the change.

Screenshots attached.

## Evidence

Before:
- 100% baseline: before-100
- Zoom out, 3 notches → 10% (model a dot): before-out
- Zoom in, 5 notches → 800%: before-in

After (this branch):
- 6 notches out → ~69%, full model still visible: after-out
- 6 notches back in → ~100%: after-in

Repro: Playwright + headless Chromium, one wheel notch = deltaY 120, against the exported STEP model (the 150×150×50 mm DGX Spark stand). Steps: load the page, wheel 6x out, 6x in, screenshot at each limit. Same script run before and after the change.


earthtojake:
Thanks for this — the bug is real and I reproduced it exactly. But I can't recommend merging as-is: the fix overshoots by roughly 8x, and it re-tunes around the cause rather than removing it.

## The bug reproduces, and it is display-dependent

I measured the toolbar zoom readout while dispatching wheel events at the canvas — 8 notches of `deltaY: 100` on the `motorcycle_shock_absorber` assembly:

| | `develop` today | this PR |
|---|---|---|
| Mouse, **dPR 1** | `100 → 60 → 36 → 21 → 13 → 10` — **−40%/notch, floor at notch 5** | `100 → 95 → 90 → 86 → 81 → 77 → 74 → 70 → 66` (−5%/notch) |
| Mouse, **dPR 2** | `100 → 77 → 60 → 46 → 36 → 28 → 21 → 17 → 13` (−23%/notch) | `100 → 97 → 95 → 93 → 90 → 88 → 86 → 84 → 81` (−3%/notch) |
| Trackpad, 20 events, dPR 2 | `100 → 75` | `100 → 96` |

Your report is confirmed by the top-left cell. But notice the two `develop` rows disagree with each other: **the same mouse zooms about 2x faster on a 1x display than on a Retina one.** That is the actual defect.

`OrbitControls.getZoomScale` is `0.95 ^ (zoomSpeed * |delta| / (100 * floor(devicePixelRatio)))`. It divides by the display DPI, and `getWheelZoomSpeed` only corrects for *changes* in that bucket (browser zoom) against `initialWheelPixelRatioBucket`, never for its absolute value. The constants were tuned on a Retina machine and land twice as hot on a 1x display — very likely what you are on.

## Two problems with the diff as written

**1. It changes normal zoom speed by ~8x.** On a Retina trackpad, 20 scroll events now move the zoom 4% (`100 → 96`); reframing a model becomes a grind. As a cure for "hits the floor in a few notches", −3%/notch overshoots — 100% to 10% would take about 45 notches.

**2. The stated reasoning cites a constant a mouse notch never uses.** The comment explains `DEFAULT_ZOOM_SPEED` (4.5) as what "made a single mouse notch move the camera ~3x farther". `handleWheel` in `useViewerRuntime.js` overwrites `controls.zoomSpeed` on *every* wheel event with `ACCELERATED_WHEEL_ZOOM_SPEED` (10), or `TRACKPAD_PINCH_ZOOM_SPEED` (14) for trackpad-like deltas — 4.5 is only the resting baseline, which touch pinch uses. So dropping it 4.5 → 1.0 slows touch pinch to fix a wheel bug.

## Suggested alternative

Cancel the DPI instead of re-tuning around it, in `useViewerRuntime.js`:

```js
const getWheelZoomSpeed = (baseZoomSpeed) => {
  // OrbitControls divides wheel deltas by a floored devicePixelRatio internally, so the same
  // notch zoomed ~2x farther on a 1x display than on a Retina one. Multiplying by that same
  // bucket cancels the division outright: one notch is one step everywhere, and because the
  // bucket is read live this also absorbs browser-zoom changes.
  return baseZoomSpeed * getOrbitControlsPixelRatioBucket();
};
```

Speeds then read as literal per-notch steps (`0.95 ^ speed` at a standard `|delta| = 100`), so `ACCELERATED_WHEEL_ZOOM_SPEED = 2.5` and `TRACKPAD_PINCH_ZOOM_SPEED = 7`. `DEFAULT_ZOOM_SPEED` and both `COARSE_POINTER_*` constants stay untouched — they are not part of this bug.

Measured with the same harness:

| | dPR 1 | dPR 2 |
|---|---|---|
| Mouse, 8 notches | `100 → 88 → 77 → 68 → 60 → 53 → 46 → 41 → 36` | **identical** |
| Trackpad, 20 events | `100 → 75` | **identical** |

The two displays now agree, which neither `develop` nor this PR achieves. −12%/notch is brisk without running away, and the trackpad curve matches `develop`'s Retina behaviour exactly, so anyone already happy with it feels no change at all.

Happy to push that as a separate PR crediting you for the find, or you are very welcome to take it into this branch — your call.

One housekeeping note: the five PNGs under `evidence/` should not ship in the merge commit. Repo convention (`AGENTS.md`) keeps durable artifacts under `models/`; screenshots like these belong in the PR description instead.

earthtojake:
Investigating this, I've only tested on my macbook pro trackpad which has reasonable zoom level. Seeing if there are any packages online that handle cross-platform zoom.

earthtojake:
Thanks for reporting this — you found a real bug, and chasing it down turned up a bigger cause than the constants.

It is an upstream three.js issue, fixed in **r161**; we were pinned to 0.160.0, one release short. r160's `getZoomScale` divides the wheel delta by `floor(devicePixelRatio)` and ignores `deltaMode`, so the same mouse notch produced anything from −40% to −2% depending on OS, browser and display — and below 1x device pixel ratio (browser zoomed out) `0.75 | 0` is `0`, so the delta was divided by zero and the wheel stopped working entirely. That last case is why re-tuning the constants could not be the whole answer.

I have opened #295 with the upgrade: it makes every mouse configuration land on the same −12% step, revives the zoomed-out case, and leaves the macOS trackpad feel untouched. Your report is what surfaced it.

Closing this in favour of #295 is up to @earthtojake — happy either way, and thanks again.

earthtojake:
#295 is merged, so closing this one — the fix that landed is the three.js r161 upgrade rather than the constant change, but it started here and would not have been found without your report. Thanks for taking the time to file it with reproduction steps.

PR #294, opened 2026-08-18
fix: npm start fails for the shipped CAD Viewer runtime

jarvis959:
## Problem

The shipped CAD Viewer runtime (bundled as skills/cad-viewer/scripts/viewer) has a start script that does not work:

- package.json says `npm start` runs `node scripts/start-viewer.mjs`, but build_runtime in the bundler never copies scripts/, so the file is missing from the bundle and npm start dies with module-not-found.
- start-viewer.mjs also imported two dev-only helpers (cad-python.mjs, directoryRoot.mjs) that are likewise not in the runtime.

Separately, start_viewer.py's port probe used connect() to test if a port was free. On Windows a closed local port times out instead of refusing the connection, so the probe reported every port as already in use and the server never started. Even when the launcher was present, npm start exited 1.

## Changes

- bundle-cad-viewer.sh: copy viewer/scripts/ into the runtime bundle.
- start-viewer.mjs: make it dependency-free by inlining the python-discovery and directory-root logic it used to get from the dev-only helpers, so the one file works from a source checkout and from the bundle.
- start_viewer.py: port_is_free now binds the socket (ground truth) instead of connecting; this works on every OS where connect-based probing was unreliable.

## Verification

On Windows, with the CAD venv python set via CAD_PYTHON and PYTHONPATH cleared:

```
$ npm start -- --host 127.0.0.1 --port 4613
Starting CAD Viewer at http://127.0.0.1:4613/
CAD Viewer URL: http://127.0.0.1:4613/
```

and the bundled app serves at that URL (checked / and /__cad/server). Also verified port_is_free directly: free ports return true, an in-tree bind returns false.

### DIFF
--- a/scripts/bundle/skills/bundle-cad-viewer.sh
+++ b/scripts/bundle/skills/bundle-cad-viewer.sh
@@ -432,6 +432,13 @@ build_runtime() {
   # tests/test_worker.py not).
   sync_dir "$VIEWER_DIR/server_py" "$target_dir/server_py"
 
+  # Node launcher (scripts/start-viewer.mjs) powers `npm start`; it is
+  # dependency-free so it works inside the bundled runtime without the dev-only
+  # helper modules (cad-python.mjs, directoryRoot.mjs).
+  if [ -d "$VIEWER_DIR/scripts" ]; then
+    sync_dir "$VIEWER_DIR/scripts" "$target_dir/scripts"
+  fi
+
   write_runtime_package_json "$target_dir"
   write_runtime_gitignore "$target_dir"
   write_runtime_requirements "$target_dir"
--- a/viewer/scripts/start-viewer.mjs
+++ b/viewer/scripts/start-viewer.mjs
@@ -1,29 +1,68 @@
 #!/usr/bin/env node
-// Thin shim: launch the Python CAD Viewer launcher (server_py.start_viewer)
-// using the project's discovered venv Python + cadgen PYTHONPATH, so `npm start`
-// can run the Python backend without the caller knowing the venv path. Reuses the
-// proven cadPythonExecutable/cadPythonEnv discovery from the STEP pipeline.
+// Self-contained `npm start` for the CAD Viewer runtime.
+//
+// Run `npm start` where the viewer runtime is installed (source checkout or the
+// bundled skill copy under scripts/viewer) and it boots the Python backend. The
+// script is intentionally dependency-free so it ships inside the bundled skill
+// (scripts/ is copied by build_runtime in scripts/bundle/skills/bundle-cad-viewer.sh);
+// the dev-only helpers it used to import (cad-python.mjs, directoryRoot.mjs) are
+// not part of the runtime distribution.
 import { spawn } from "node:child_process";
+import { existsSync } from "node:fs";
 import path from "node:path";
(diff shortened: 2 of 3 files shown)


earthtojake:
Three bugs here share one cause: CI never runs the thing we ship.

The bundler stopped copying `viewer/scripts/`, so `npm start` in the bundled runtime has not worked since. The runtime's `package.json` declares `"start": "node scripts/start-viewer.mjs"` and no test ever ran it.

The durable fix is a smoke test. Build the bundle in CI, run `npm start` from it, curl `/` and `/__cad/server`, kill it, and do the same on `windows-latest`. That covers this bug and the port probe below. The repo has the pattern for Python in `test-installed.sh`.

Two design points for the code.

**Drop the port probe.** `port_is_free` asks whether a bind would succeed, then hands the port to a child process that binds for real. The check and the bind race each other, and they disagree about socket options: `ThreadingHTTPServer` sets `allow_reuse_address`, a bare probe socket does not. I measured it. After you stop the Viewer, the bind probe calls the port busy while the backend can still claim it, so restarting on the same port fails.

Let the bind report instead. Wrap `ThreadingHTTPServer(...)` at `server.py:366` in `try/except OSError`, print the "Port N in use" message there, exit non-zero, and have `start_viewer.py` return the child's code. The race disappears, the real listener's options decide, and nothing connects to a closed port, which is what broke Windows.

**Keep the helpers shared.** `vite.config.mjs` imports `cad-python.mjs` and `directoryRoot.mjs`. A second inlined copy lets the dev server and the launcher drift on Python discovery. `AGENTS.md` puts shared runtime helpers under `packages/` and vendors them into each runtime.

On the current diff: `findPython()` searches `viewer/.venv` alone, and this repo keeps its venv at the root, so a checkout falls back to system `python3`. `buildPythonPath()` drops the `viewerRoot` entry that lets `python -m server_py.start_viewer` resolve `server_py`. Running both branches on macOS with no `CAD_PYTHON` set: `develop` answers HTTP 200, this branch exits with `No module named 'server_py'`.

earthtojake:
Closing this one, with thanks: you found a real bug that had shipped broken in every release since 0.4.0, and your diagnosis of the `cadjs` resolution failure was the part that made the fix work.

#297 landed that fix and went out in 0.4.19. It takes a different route on both halves, which is why I did not merge this directly: it copies the three launcher files by name rather than the whole `scripts/` directory, and it keeps `directoryRoot.mjs` shared with `vite.config.mjs` by importing the vendored file by path instead of inlining a second copy. You are credited as `Co-Authored-By` on that commit.

**The port probe is still worth fixing and nobody has done it.** `port_is_free` is unchanged on `develop`, so the Windows behaviour you identified is still there. If you want to open a narrow PR for just that, it would be very welcome. My suggestion from [this thread](https://github.com/earthtojake/text-to-cad/pull/294#issuecomment-5336309233) is to drop the probe rather than swap it: wrap the `ThreadingHTTPServer(...)` construction at `server.py:366` in `try/except OSError`, print the "Port N in use" message there, and have `start_viewer.py` return the child's exit code. That removes the Windows connect-timeout, the check-then-bind race, and the `SO_REUSEADDR` mismatch together, and it needs no probe socket at all.

PR #301, opened 2026-08-19
label_refs: accept the occurrenceId row spelling, and run aliasCases in Python

NgoQuocViet2001:
## What changed

`buildLabelAliasMap` reads `row.id || row.occurrenceId`. That matches the render package — `cadjs/common/cadScene.js` reads `part.id || part.occurrenceId` in five places and `cadScene.test.js` fixtures key their parts that way. `build_label_aliases` read only `id`, so any row carrying the other spelling was dropped before it could claim an alias, and a label the viewer resolves had no alias in the CLI.

Differential run over the same row sets, before this change:

```
rows: [{"occurrenceId":"o1.1","name":"eye_shank"}, {"id":"o1.2","name":"pressure_tube"}]
  py: {"aliases": {"spaced": ...}}                  -> eye_shank missing
  js: {"aliases": {"eye_shank":"o1.1", ...}}
```

## Why the fixture did not catch it

`cadRefs.parity.json` says every case must hold in both languages, and it carries `aliasCases` — but nothing on the Python side ever read them. `ParityFixtureIsUsableTest` only asserts the key is non-empty, and the imports in `test_cad_ref_syntax_parity.py` stop at the selector grammar, so `build_label_aliases` was never run against the fixture at all. The alias builders could drift with the suite still green, which is exactly what happened.

This adds `AliasParityTest`, so the fixture now does for the alias map what it already does for selectors, plus the case that covers the row spelling.

## Test plan

- Ran: `PYTHONPATH=".;packages/cadgen/src" python -m unittest tests.python.packages.cadgen.test_cad_ref_syntax_parity tests.python.packages.cadgen.test_label_refs` — 35 passed
- Ran: `node --test src/lib/cadRefs.test.js` in `packages/cadjs` — 13 passed
- Checked: with the `label_refs.py` change reverted, the new alias case fails with `{'eye_shank': 'o1.1', 'pressure_tube': 'o1.2'} != {'pressure_tube': 'o1.2'}` — so the test is pinned to the behaviour, not to the fixture
- Checked: a differential run over nine row sets (duplicate names, authored `_1` collisions, nested occurrence ids, blank ids, mixed case, unaliasable names) now agrees between the two implementations on all nine
- Not run: `test_assembly_selector_refs` (no `build123d` on this checkout — it fails to import on a clean `develop` here too)

### DIFF
--- a/packages/cadgen/src/cadgen/label_refs.py
+++ b/packages/cadgen/src/cadgen/label_refs.py
@@ -71,7 +71,12 @@ def build_label_aliases(rows: Iterable[Mapping[str, Any]]) -> dict[str, Any]:
     """
     ordered: list[tuple[str, str]] = []
     for row in rows:
-        occurrence_id = str(row.get("id") or "").strip()
+        # `occurrenceId` is the spelling the render package uses for the same
+        # field (cadjs/common/cadScene.js reads `part.id || part.occurrenceId`
+        # throughout), and buildLabelAliasMap accepts both. Reading only `id`
+        # here dropped those rows, so a label the viewer resolves had no alias
+        # in the CLI.
+        occurrence_id = str(row.get("id") or row.get("occurrenceId") or "").strip()
         if not occurrence_id:
             continue
         candidate = _label_candidate(row.get("name"))
--- a/packages/cadjs/src/lib/cadRefs.parity.json
+++ b/packages/cadjs/src/lib/cadRefs.parity.json
@@ -379,6 +379,25 @@
         "pressure_tube": "o1.1.1"
       },
       "ambiguous": {}
+    },
+    {
+      "phase": 1,
+      "why": "a row may carry the render package's `occurrenceId` spelling instead of `id`",
+      "rows": [
+        {
+          "occurrenceId": "o1.1",
+          "name": "eye_shank"
+        },
+        {
+          "id": "o1.2",
+          "name": "pressure_tube"
+        }
+      ],
+      "aliases": {
+        "eye_shank": "o1.1",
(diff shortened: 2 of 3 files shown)


earthtojake:
The alias parity test is the strong half of this, and it closes a gap I left. `cadRefs.parity.json` has carried `aliasCases` since it was written, and nothing on the Python side read them. `ParityFixtureIsUsableTest` checked that the key held something and stopped there, so the two alias builders could drift apart with the suite green. `AliasParityTest` ends that. I reverted the `label_refs.py` change and the new case fails, so the test holds the behaviour rather than the fixture.

I could not reproduce the symptom in the description. Three things point the other way:

- `buildLabelAliasMap` has no production callers in JS. Only tests reach it.
- Every Python caller supplies `id`. `attach_label_aliases` reads `index.occurrences`, and `interference.py` constructs `{"id": occurrence.ref, "name": ...}`.
- `assembly.json` occurrence rows carry `id`, `name`, `component`, `transform`, `color`. No `occurrenceId`.

So a row with that spelling reaches neither builder today, and the differential in the description runs on input the pipeline does not produce. The divergence is real as a contract question and dormant as a defect.

That is still worth taking. Parity is what the fixture promises, `row.get("id") or row.get("occurrenceId")` matches the JS precedence exactly, and the cost is one `or`. I would reframe the description around the test gap, which is the part that protects future work.

Two mechanics for you. GitHub refused to update this branch (`user doesn't have permission to update head repository`), and `develop` requires branches to be current, so I opened #302 carrying your commit with `develop` merged in and your authorship intact. CI passes there. Both are on hold pending a call from @earthtojake, so nothing is merged yet.

PR #304, opened 2026-08-20
viewer: enforce root containment, fix symlink escapes, and guard scanner loops

warun7:
## Summary

Hardens CAD Viewer root containment, prevents symlink directory escapes and recursion cycles, and adds file manager reveal support:

1. **Root Directory Containment & Validation**:
   - File-serving routes (`/__cad/asset`, `/__cad/download`, `/__cad/reveal`, and legacy Referer routes) now require a non-empty `dir=` query parameter. Requests without an active directory receive `400 Bad Request` (`{"error": "Missing directory"}`).
   - Backend containment validation in `asset_path_for_file_ref` runs unconditionally, rejecting paths outside the served root.
   - Fixed dot-directory filtering order so files inside hidden folders under the served root return 404.

2. **Symlink Escape Prevention & Scanner Cycle Guards**:
   - `scanner.path_is_inside` compares `os.path.realpath` for both the requested file and root directory so external symlinks cannot bypass containment.
   - `_collect_cad_source_files` tracks visited `realpath` directories and enforces a recursion depth limit (`_SCAN_MAX_DEPTH = 64`) to prevent `RecursionError` on cyclic directory symlinks.
   - Symlinks escaping the served root are excluded from catalog indexing.

3. **Explicit 404 for Unrecognized CAD Endpoints**:
   - Unhandled `/__cad/*` requests return `404 Not Found` JSON instead of falling through to the SPA `index.html`.

4. **OS Reveal Endpoint & Client URL Integration**:
   - Added `POST /__cad/reveal` with platform-specific commands (`open -R`, `explorer /select,`, `xdg-open`) bound by root containment.
   - Updated client URL builders in `fileAccessAssets.js` and `modelExport.js` to pass `dir=`.

## Testing

- Added `viewer/server_py/tests/test_server_routes.py` covering route validation, containment, hidden directories, symlink escapes, and reveal handling.
- Added `viewer/server_py/tests/test_scanner.py` covering symlink resolution, loop termination, and escape filtering.
- Ran all Python tests: 195/195 passed.
- Ran all JavaScript tests: 366/366 passed.
- Verified viewer self-containment and bundle checks.


earthtojake:
Sorry for the delayed response, and thanks for the contributions so far.

The scan loop guard is the part we want. A `ln -s . loop` inside a served folder
recurses to RecursionError on every scan today, and your visited-realpath set with a
depth cap fixes it while keeping the walk deterministic. The macOS aliasing fix rides
along for free, since `/var` and `/private/var` now compare clean.

We can't take the symlink refusal, because symlinked model directories are a feature
of this tool. The develop checkout itself runs a symlink layout, and pointing a link
at a shared parts library outside the served folder is the normal way to bring
external content into a workspace. Under this change those folders vanish from the
catalog and their files 403 at serve time. The refusal also buys no boundary on
develop, where the viewer opens any absolute directory named in the URL: anyone who
could follow a symlink out of a root can request the target directly, so a link out
of the root adds no reach the URL didn't already grant.

One architecture note before you rework it. The viewer is moving to one served root
per instance, fixed at startup (release/0.5.0, PR #273 and follow-ups). Containment
becomes unconditional there, and symlinks are followed and served on purpose, the
same posture Jupyter takes, pinned by test on that branch. A parallel containment
mechanism on develop would diverge from that and get replaced at the rebase.

Suggested re-scope: keep the cycle guard, the depth cap, the deterministic ordering,
and the tests covering them. Use realpath so aliases compare equal, never to refuse.
Drop the escape refusal in scanner and backend. That version is a quick merge.

If you want to chat further, message me on Discord: https://discord.gg/5FGB9DwJYU

warun7:
Re-scoped in b0f59af3, thanks for the review.

- Scanner follows directory symlinks again (visited-realpath cycle guard + depth cap unchanged, ordering deterministic).
- `path_is_inside` now accepts a path contained **lexically or** after resolution: realpath exists purely for alias equality (`/var` ≡ `/private/var`), never to refuse. A link to a parts library outside the open folder scans and serves as before.
- Escape-refusal tests inverted into keep-the-feature tests: a link pointing outside the root scans into the catalog and serves (200); alias equality is pinned; the loop/depth tests stand.

The rest of the PR (missing-`dir` → 400 on file routes, 404 JSON for unrecognized `/__cad/*`, reveal route) is untouched.

PR #307, opened 2026-08-21
fix(dfam-check): move __main__ guard to end of test file

knewnothing-git:
Follow-up from the review on #265.

The `if __name__ == "__main__": unittest.main()` block sat mid-file, with `MultiBodyWallTest`, `ResilienceTest` and `ScaleHintTest` defined below it. A direct `python test_dfam_tool.py` therefore collected only the 11 classes above the guard and still printed OK — and the 7 it skipped included the `per_body` `p05_mm` coverage from the last round.

CI runs `python -m unittest <paths>`, which imports the module, so all 18 ran there and nothing slipped through on the merged PR.

Moved the guard to the end of the file. Verified before and after:

| path | before | after |
|---|---|---|
| `python test_dfam_tool.py` | 11 tests, OK | 18 tests, OK |
| `python -m unittest <path>` | 18 tests, OK | 18 tests, OK |
| `pytest` | 18 passed | 18 passed |

Self-containment gate still passes. One file, +4/-3.

### DIFF
--- a/tests/python/skills/dfam-check/test_dfam_tool.py
+++ b/tests/python/skills/dfam-check/test_dfam_tool.py
@@ -212,9 +212,6 @@ class CliTest(unittest.TestCase):
         self.assertEqual(len(payload["orientations"]["candidates"]), 6)
 
 
-if __name__ == "__main__":
-    unittest.main()
-
 
 class MultiBodyWallTest(unittest.TestCase):
     """A mating clearance is not a wall.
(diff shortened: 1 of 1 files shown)


knewnothing-git:
Ah — that's on me. I appended those classes without checking where the guard sat, so they landed below it.

Reproduced it before fixing: direct run gave 11 and still printed OK, and the 7 it skipped included the p05_mm coverage from the last round. Moved the guard to the end — direct run now collects all 18, unittest and pytest paths unchanged. PR #307.

Thanks for three rounds of review. The clean-venv runs caught things I'd have shipped otherwise.

PR #325, opened 2026-08-24
viewer: per-parameter reset for number parameters

earthtojake:
The Parameters tab had one Reset, and it reset everything. Changing one slider to try
something and then wanting only that one back meant either remembering its default or
resetting the lot and redoing the rest.

**Number** and **colour** parameters now each carry their own reset beside their
control: icon-only, the same `RotateCcw` the tab's other resets already use, styled
`outline` so it carries the same border and surface as the value inputs, dropdowns and
swatches it sits in a column with. The tooltip names the value it will restore
(`Reset to 0.00 deg`, `Reset to #00a7ff`).

It stays live even when the parameter is already on its default. Disabling it there was
tried and dropped: a control greying out normally means the app is saying something is
unavailable, so a reset that dims whenever the value looks right invites the reading
that the row is broken rather than that it is already where it started.

The bulk Reset is unchanged and still resets everything.

## Why these two types

Counting declarations across `models/`:

| Type | Declarations | Files |
|---|---|---|
| `number` | 297 | — |
| `color` | 109 | 44 |
| `boolean` | 28 | — |
| `select` | 15 | 14 |
| `string` | 0 | 0 |
| `button` | 0 | 0 |

Colour is the value you are least able to get back by eye: the swatch shows where you
are with no hint of where you started, and unlike a number there is no typing your way
home.

`string` and `button` have **zero** declarations anywhere, so a reset for them would be
built on a guess. `boolean` and `select` are real but weak — both show their state in
the control itself and both pick from a small finite set, so returning to the default
is a click rather than a recovery.

## Verified in the browser

`planetary_gear_assembly.step.py` (STEP, numbers) and `planetary-gear.implicit.js`
(implicit, six colours and ten numbers — the same section serves both file sheets):

- 16 resets render, every one sharing a single right edge (1272px) and a single height
  (28px) with the inputs beside them
- each colour's tooltip carries its own declared default
- Drive `0.00 → 723 deg → reset → 0.00 deg`, returning the geometry to its original
  pose and leaving Explode untouched
- Ring color `#00A7FF → #5D727E → reset → #00A7FF`
- no console errors

371 viewer tests pass.

### DIFF
--- a/viewer/src/client/components/workbench/ParameterControlsSection.js
+++ b/viewer/src/client/components/workbench/ParameterControlsSection.js
@@ -38,6 +38,44 @@ function formatControlNumber(value) {
   return numericValue.toFixed(2);
 }
 
+// A parameter's own way back to its declared default, beside the control it resets.
+//
+// Outline rather than ghost: it sits in a column of value inputs, dropdowns and colour
+// swatches, and `outline` carries the same border and surface those do, at the h-7
+// every control in the sheet shares.
+//
+// Live even at the default. A control greying out normally means the app is saying
+// something is unavailable, so a reset that dims whenever the value looks right invites
+// the reading that the row is broken rather than that it is already where it started.
+// Resetting to where you already are costs nothing.
+function ParameterResetButton({ label, title, enabled, onReset }) {
+  return (
+    <Button
+      type="button"
+      variant="outline"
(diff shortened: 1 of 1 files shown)


earthtojake:
Closing — not needed. The bulk Reset covers the case well enough on its own.

PR #326, opened 2026-08-24
fix(dxf): read the arc angle keys the parsers actually emit

NgoQuocViet2001:
## The bug

Every DXF `ARC` renders as a **complete circle** in the drawing view.

`sampleArc` reads `arc.startAngle` / `arc.endAngle`. Neither producer emits those keys — `parseDxf.js` and `drawing_render.py` both emit `startAngleDeg` and `sweepAngleDeg`. So `start` and `end` both fall back to `0`, `sweep` comes out `0`, and the wrap immediately below turns that into a full turn:

```js
let sweep = ((end - start) * Math.PI) / 180;   // 0
if (sweep <= 0) {
  sweep += Math.PI * 2;                        // 360°
}
```

A 90° arc at the origin, radius 10 — the payload `parseDxf` produces for `code 50 = 0`, `code 51 = 90`:

```
before   48 segments, last point back on the first, x spans -10.00 .. 10.00
after    12 segments, x spans   0.00 .. 10.00
```

Bounds are wrong too, since `drawingLineBounds` measures the sampled points — a quarter arc reports the full circle's extent.

## The fix

Prefer `startAngleDeg`, and derive the sweep from `sweepAngleDeg`, which both producers guarantee non-zero in `(0, 360]`.

The end-angle path stays as the fallback, so `sampleCircle`'s `{ startAngle: 0, endAngle: 360 }` call and any legacy payload keep working unchanged — including the `sweep <= 0` wrap that makes a bare arc with no angles still close into a circle.

## Tests

`packages/cadjs` had no tests, and `scripts/run-tests.mjs` exits with *"No cadjs tests found."* — so this adds the first ones, at the path that runner already collects.

They assert the sampled geometry rather than the key names, which is what actually drifted here:

| test | asserts |
|---|---|
| a quarter arc in parser keys stays a quarter arc | x stays ≥ 0 |
| a full circle in parser keys still closes | x spans −10..10 |
| the legacy start/end angle shape is still honoured | x stays ≥ 0 |
| an arc with no angles at all still yields a closed circle | x spans −10..10 |

```
after   ℹ pass 4   ℹ fail 0
before  ℹ pass 3   ℹ fail 1   ✖ a quarter arc in parser keys stays a quarter arc
```

Only the case that was broken fails on `main`; the three control cases pass either way, which is what makes the fallback safe.

### DIFF
--- a/skills/cad-viewer/scripts/viewer/packages/cadjs/src/lib/dxf/buildDrawingLines.js
+++ b/skills/cad-viewer/scripts/viewer/packages/cadjs/src/lib/dxf/buildDrawingLines.js
@@ -38,14 +38,28 @@ function sampleArc(target, arc, elevation, requested) {
   if (!Number.isFinite(radius) || radius <= 0) {
     return;
   }
-  const start = Number(arc.startAngle ?? arc.start_angle ?? 0);
-  const end = Number(arc.endAngle ?? arc.end_angle ?? 0);
-  if (!Number.isFinite(start) || !Number.isFinite(end)) {
+  // parseDxf and drawing_render.py both emit startAngleDeg/sweepAngleDeg.
+  // Reading only startAngle/endAngle left start and end at 0 for every real
+  // arc, so sweep came out 0, the wrap below turned it into a full turn, and
+  // each arc was drawn as a complete circle.
+  const start = Number(arc.startAngleDeg ?? arc.startAngle ?? arc.start_angle ?? 0);
+  if (!Number.isFinite(start)) {
     return;
   }
   // Angles arrive in degrees, counter-clockwise, as DXF stores them.
   const startRadians = (start * Math.PI) / 180;
-  let sweep = ((end - start) * Math.PI) / 180;
+  let sweep;
+  const sweepDeg = Number(arc.sweepAngleDeg ?? arc.sweep_angle_deg);
+  if (Number.isFinite(sweepDeg) && sweepDeg !== 0) {
+    // Both producers guarantee a non-zero sweep in (0, 360].
+    sweep = (sweepDeg * Math.PI) / 180;
+  } else {
+    const end = Number(arc.endAngle ?? arc.end_angle ?? 0);
+    if (!Number.isFinite(end)) {
+      return;
+    }
+    sweep = ((end - start) * Math.PI) / 180;
+  }
   if (sweep <= 0) {
     sweep += Math.PI * 2;
   }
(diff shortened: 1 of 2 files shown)


earthtojake:
The diagnosis and the fix are both right, and I reproduced the bug before reading your
patch: feeding a real `{ startAngleDeg: 0, sweepAngleDeg: 90 }` payload through
`buildDrawingLines.js` on `develop` produces output **byte-identical** to a deliberate
full circle — 96 points, spanning -10..10 on both drawing axes. With your logic applied
it becomes 24 points spanning 0..10, `sampleCircle`'s explicit `startAngle:0,
endAngle:360` call still draws a full circle, and a bare arc with no angles still closes.
Preferring `sweepAngleDeg` over recomputing from `endAngleDeg` is the right call too,
since both producers guarantee it non-zero.

The problem is where it lands, not what it does.

**The base branch.** This targets `main`, which is publish-only — releases are cut from
`develop` and published to `main`, so a change that starts on `main` is overwritten by
the next release rather than carried forward. Repo PRs should target `develop`. It also
means CI never runs here: `test.yml` only triggers on pull requests to `develop`, which
is why this PR shows no checks.

**The file.** `skills/cad-viewer/scripts/viewer/packages/cadjs/...` is a *generated*
copy. It does not exist on `develop` at all — the publish job builds that directory from
`packages/cadjs` when it cuts a release. The blob you edited is byte-identical to
`packages/cadjs/src/lib/dxf/buildDrawingLines.js` for exactly that reason. Merged as-is,
the next release would regenerate the directory from `develop` and silently erase the
fix.

To re-file:

- base the PR on `develop`
- apply the change to `packages/cadjs/src/lib/dxf/buildDrawingLines.js`
- add your cases to the existing `packages/cadjs/src/lib/dxf/buildDrawingLines.test.js`
  rather than a new `.test.mjs` — that file is already there on `develop` (the published
  copy you branched from does not carry tests, which is why it looked absent)

Worth doing — it is a real bug and every DXF arc in the viewer is affected. Happy to
review the re-filed version.

NgoQuocViet2001:
You're right, and thanks for spelling out why rather than just closing it. I branched from the published tree and never checked whether `skills/cad-viewer/.../packages/cadjs` was generated — so the next release would have regenerated it and quietly erased the fix. That is a worse outcome than the bug.

Re-filed as #331 against `develop`: the change is in `packages/cadjs/src/lib/dxf/buildDrawingLines.js`, and the four cases are appended to the existing `buildDrawingLines.test.js` rather than a new file. `node scripts/run-tests.mjs src/lib/dxf/buildDrawingLines.test.js` → 12 passed (the existing 8 plus these 4).

The MTEXT one is re-filed as #332 the same way.

Closing this.

PR #327, opened 2026-08-24

NgoQuocViet2001:
## The bug

The paragraph-break replace carries the `i` flag:

```js
text = text.replace(/\\P/gi, "\n").replace(/\\~/g, " ");
```

`\P` is a paragraph **break**. Lowercase `\p` is paragraph **properties** — a different command that carries a payload up to a semicolon (`\pxqc;`, `\pxi-2,l2,t2;`). Because `p` is also absent from the inline-property class on the next line, the replace consumes only the two characters and leaves the rest in the text:

```
\pxqc;PART A        ->  "xqc;PART A"        (want "PART A")
\pxi-2,l2,t2;Item   ->  "xi-2,l2,t2;Item"   (want "Item")
```

Any MTEXT carrying a paragraph alignment or indent engraves the command payload. Centred titles (`\pxqc;`) and indented notes are the everyday cases, so this shows up on ordinary drawings rather than exotic ones.

## The fix

Make the break case-sensitive and add `p` to the property-run class:

```js
text = text.replace(/\\P/g, "\n").replace(/\\~/g, " ");
text = text.replace(/\\[fFhHcCtTqQwWaAp][^;]*;/g, "");
```

Order already works in favour of this: the `\P` replace runs first, so no uppercase `\P` survives to be eaten by the property run.

## Verified

```
"\pxqc;PART A"        -> "PART A"          OK
"\pxi-2,l2,t2;Item"   -> "Item"            OK
"Line1\PLine2"        -> "Line1\nLine2"    OK   (break still a newline)
"\H2.5x;BIG"          -> "BIG"             OK
"plain"               -> "plain"           OK
```

## Tests

Added `src/lib/dxf/stripMtextFormatting.test.mjs`, covering the properties command, the break, the other property runs, and plain text.

```
after   ℹ pass 4   ℹ fail 0
before  ℹ pass 3   ℹ fail 1   ✖ paragraph properties are removed whole
```

Only the broken case fails on `main` — the paragraph-break and other-property tests pass either way, which is what makes the case-sensitivity change safe.

Sits alongside the arc-sampling fix in #326; the two touch different files and can land in either order.

### DIFF
--- a/skills/cad-viewer/scripts/viewer/packages/cadjs/src/lib/dxf/parseDxf.js
+++ b/skills/cad-viewer/scripts/viewer/packages/cadjs/src/lib/dxf/parseDxf.js
@@ -679,9 +679,12 @@ const NON_GEOMETRIC_ENTITY_TYPES = new Set([
 export function stripMtextFormatting(raw) {
   let text = String(raw ?? "");
   // \P is a paragraph break, \~ a hard space; \\ and \{ \} escape literals.
-  text = text.replace(/\\P/gi, "\n").replace(/\\~/g, " ");
-  // Inline property runs: \f...; \H...; \C...; \T...; \Q...; \W...; \A...; — command up to ;
-  text = text.replace(/\\[fFhHcCtTqQwWaA][^;]*;/g, "");
+  // Case-sensitive: lowercase \p is paragraph PROPERTIES (\pxqc;), a different
+  // command. Matching it here consumed only the two characters and left its
+  // payload behind, so \pxqc;PART A engraved as "xqc;PART A".
+  text = text.replace(/\\P/g, "\n").replace(/\\~/g, " ");
+  // Inline property runs: \f...; \H...; \C...; \T...; \Q...; \W...; \A...; \p...; — command up to ;
+  text = text.replace(/\\[fFhHcCtTqQwWaAp][^;]*;/g, "");
   // Stacking \S...^...; renders as the plain parts.
   text = text.replace(/\\S([^^;]*)\^([^;]*);/g, "$1/$2");
   // Grouping braces are structure, not content.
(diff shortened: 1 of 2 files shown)


earthtojake:
The catch is correct and the reasoning is exactly right: `\P` is a paragraph break, `\p`
is paragraph properties, and matching case-insensitively consumed the two characters and
left the payload. I confirmed it against `develop` before reading the patch —
`\pxqc;PART A` currently strips to `"xqc;PART A"`.

Your fix passes every case I threw at it, including two beyond your tests: a string
carrying both a break and properties (`A\PB\pxqc;C` → `"A\nBC"`), and stacking
(`\S1^2;` → `"1/2"`) still surviving after adding `p` to the property-run class.

The problem is where it lands, not what it does.

**The base branch.** This targets `main`, which is publish-only — releases are cut from
`develop` and published to `main`, so a change that starts on `main` is overwritten by
the next release rather than carried forward. Repo PRs should target `develop`. It also
means CI never runs here: `test.yml` only triggers on pull requests to `develop`, which
is why this PR shows no checks.

**The file.** `skills/cad-viewer/scripts/viewer/packages/cadjs/...` is a *generated*
copy. It does not exist on `develop` at all — the publish job builds that directory from
`packages/cadjs` when it cuts a release. Merged as-is, the next release would regenerate
it from `develop` and silently erase the fix.

To re-file:

- base the PR on `develop`
- apply the change to `packages/cadjs/src/lib/dxf/parseDxf.js`
- add your cases to the existing `packages/cadjs/src/lib/dxf/parseDxf.test.js` rather
  than a new `.test.mjs` — that file is already there on `develop` (the published copy
  you branched from does not carry tests, which is why it looked absent)

Same note applies to #326. Both are real bugs and both fixes are sound; they just need
to land on the source rather than the published copy.

NgoQuocViet2001:
Same mistake as #326 — I branched from the published tree and edited a generated copy, so a release would have regenerated it away. Thanks for catching both in one pass.

Re-filed as #332 against `develop`: the regex change is in `packages/cadjs/src/lib/dxf/parseDxf.js`, with three cases appended to the existing `parseDxf.test.js` instead of a new file. `node scripts/run-tests.mjs src/lib/dxf/parseDxf.test.js` → 15 passed, and 14 passed / 1 failed with the source reverted, so the new case is pinned to the behaviour.

The arc fix is #331. Closing this.

PR #329, opened 2026-08-24
cad: document inspect interfere in the required workflow

13thrule:
## Summary

`scripts/inspect interfere` exists and works correctly, but is not mentioned anywhere in `skills/cad/SKILL.md`'s required workflow or `references/inspection-and-validation.md`. Step 8 of the required workflow names `measure`/`align`/`frame`/`diff` as the targeted-check options and omits `interfere`; the doc's only other match for the word "interfere" is an unrelated visual-review display-mode description (spotting overlap by eye in a snapshot), not the command.

## Why this matters

Neither `refs --facts` nor `validate` test one positioned part against another — each solid can be individually well-formed (closed, positive volume, correct topology) while still occupying the same space as a different part. Concretely, all of these pass both checks clean and only show up under `interfere`:

- two gears with a tooth-phase error
- a bracket clamped through the case it's supposed to clear
- a joint rotated into its own previous link

Working through a small assembly build this week, `inspect interfere` was the single most valuable check for catching real geometry bugs — every one of the above happened and was only caught because someone thought to run it, not because the workflow said to.

## Fix

- `SKILL.md` step 8: add `interfere` alongside the existing baseline (`refs --facts --planes --positioning`, `validate`) for any assembly with more than one positioned part, with a note to re-run it after any positioning or joint-angle change, not just once at the pose the assembly was built at.
- `inspection-and-validation.md`: add `interfere` as step 4 of the "Validation sequence" list, and a new subsection ("Neither `refs` nor `validate` checks part-vs-part interference") matching the doc's existing pattern (e.g. the `refs --facts` "ok" is not a geometry claim" subsection), with the same real-world failure examples.
- Add an "Interference" line to the validation report structure so it's part of what gets reported, not just what gets run.

Documentation-only change, no code touched.


earthtojake:
I think this is covered in 330

earthtojake:
Closing in favour of #330, which contains this PR's changes verbatim plus the range-of-motion work. See the discussion there.

PR #330, opened 2026-08-24
cad: require range-of-motion interference checks for moving joints

13thrule:
## Summary

`inspect interfere` only ever gets run at whatever pose `gen_step()` happens to default to. Nothing in the required workflow ties a defined joint (a `RevoluteJoint`/`LinearJoint`/`CylindricalJoint`/`BallJoint`, or a raw `Location`/`Rotation` driven by an angle/position parameter) to actually checking interference across its stated range — so an assembly that is kinematically incapable of moving at all can still pass every existing validation step by never being tested away from rest.

This closes that gap end to end:

- `cad-brief.md` now asks for a moving joint's stated range while writing the brief, so the range exists as a captured fact before coding starts.
- `SKILL.md`'s plan step (5) requires identifying that range up front, and the validate step (8) requires sweeping it with `interfere` as part of the *same* validation pass — not only reactively after some later manual change, which is the only trigger the existing text gave.
- `positioning.md` cross-references the requirement right where joint types are chosen, since that's where an agent is already thinking about motion.
- `inspection-and-validation.md` spells out the concrete steps (identify the range, generate/pose at representative points including both limits, run `interfere` at each) and what counts as adequate coverage, plus updates the validation-report structure to ask for the poses swept, not just a clash count.

**Depends on #329** — `interfere` isn't mentioned anywhere in the `cad` skill's docs on `develop` yet (confirmed: zero hits for "interfere" across `SKILL.md`/`references/*.md` pre-#329), so this branch is stacked on top of that one rather than re-deriving the same base documentation independently. The diff here will look smaller once #329 lands first; until then this PR's diff includes #329's commit too.

## How this was found

Not synthetic — found by hand while building a real two-servo pan/tilt sensor turret. A yoke bracket was rigidly bolted to a servo horn at one end (torque only, by design) and pinned into a support bearing pressed into the stationary base plate at the other end, 40mm away on a different XY axis (meant to be "coaxial" per the design intent, but built on a separate axis by mistake). `inspect interfere` passed cleanly at the generated rest pose — every existing required check was green.

The assembly cannot move at all. Tracing it by hand: if the servo horn rotates by θ, the whole rigid yoke rotates by θ around the horn's own axis, so the far end (the bearing pin) sweeps a circle of radius 40mm around that axis. At θ=15° the far end needs to be ~10mm from where the fixed bearing actually sits; at θ=90° it needs to be ~57mm away. The pin is captured inside a fixed bearing bore, so none of that is physically possible — the mechanism is locked at whatever single pose it happened to be generated at.

No existing validation step would ever have caught this, because nothing required checking a second pose, and the generator never even exposed the joint angle as something that *could* be swept.

## Test plan

- [x] Doc-only change (four `references/*.md` + `SKILL.md` files); no code paths touched.
- [x] Ran `tests/python/global/test_skill_self_containment.py` and `tests/python/global/test_plugin_manifests.py` — both pass, confirming the skill directory structure and manifests are unaffected.
- [x] Read through the full resulting `inspection-and-validation.md`, `positioning.md`, `cad-brief.md`, and `SKILL.md` diffs for internal consistency and cross-reference correctness.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_017KwJ8FmG5KAfPbN4kyzL8a


earthtojake:
@13thrule The range-of-motion requirement in this PR needs a tool that does not
exist yet. The runner calls `gen_step()` with no arguments, so nothing can pose a
generated assembly at a non-default joint value, and the `--params` that exists on
`snapshot` drives the render-time sidecar, which `interfere` never sees. "Run
`interfere` across the range" has no executable form today.

Would you build the missing piece first? The scope is below. The plumbing exists,
so the change is small and well bounded.

## Feature: `inspect interfere --param NAME=VALUE`

Pose a generated assembly by passing keyword arguments to its `gen_step()`, then
run the existing interference check on that transient scene. One invocation per
pose:

```bash
scripts/inspect interfere models/arm/arm.step.py --param elbow_deg=0
scripts/inspect interfere models/arm/arm.step.py --param elbow_deg=90 --refs o1.2,o1.5
```

Generators opt in by declaring keyword parameters with defaults:

```python
def gen_step(elbow_deg: float = 0.0): ...
```

## Contract

- `--param NAME=VALUE`, repeatable. Generated entries (`.step.py`) only. An
  imported STEP errors, and the message has to head off a known confusion:
  imported STEPs do have sidecar parameters (the Viewer discovers
  `<name>.step.js` beside the file; `snapshot --params-path` names one by path),
  but a sidecar executes in the JS render runtime and poses the rendered scene.
  `interfere` checks the brep, which a sidecar never touches. Suggested text:
  "imported STEP: --param drives gen_step() arguments, which this entry does not
  have; its .step.js sidecar poses the render only and cannot pose the
  interference check".
- `NAME` must be a parameter `gen_step()` declares. Unknown name: error listing
  the declared names. A generator that declares none: say so.
- `VALUE` coerces from the type of the parameter's default (`int`, `float`,
  `bool` as `true`/`false`, `str`). On failure, the error names the parameter,
  the expected type, and the given text.
- Without `--param`, behaviour is byte-identical to today.
- Echo the applied params in the JSON payload and the text header, so each line
  of a sweep is self-describing.
- The flag is `--param`, singular. `snapshot --params` means sidecar JSON values,
  a different system that `interfere` cannot see, and the two spellings should
  stay distinct.

## Why it's small

`inspect_interference` (`cadgen/interference.py:277`) rebuilds the scene
in-process on every invocation via `_resolve_spec_and_scene`
(`cadgen/step_export_target.py:87`), under `lock_intent="generate"` (line 130),
the non-write lock. Nothing on this path touches the render package. The scene
cache keys on a STEP file hash and serves imported-STEP loads only, so a posed
generated scene cannot poison it. The change threads a dict through three
signatures down to `generation_runner.py`'s `raw_payload = generator()` and
validates it at that boundary.

Suggested path:

1. `skills/cad/scripts/inspect/inspect_refs/cli.py` (~line 148): add `--param`,
   parse pairs into a dict.
2. `inspect_interference(entry, ..., generator_params=None)`: thread through.
3. `_resolve_spec_and_scene(..., generator_params=None)`: pass down; raise for an
   imported STEP with params; keep `lock_intent="generate"`.
4. `run_script_generator`: accept the dict, validate against
   `inspect.signature(generator)` (names and coercion) before calling, then
   `generator(**params)`.

## Invariants review will check

1. A `--param` run writes nothing: no render package, no scene cache, no STEP.
2. Unknown or uncoercible params fail closed.
3. Zero behaviour change when the flag is absent.
4. Each invocation is independent, with no cross-invocation state, so a sweep is
   a plain shell loop.

## Tests

- A fixture generator `gen_step(angle: float = 0.0)` building two bars that clash
  only at `angle=90`: the default passes, `--param angle=90` reports the pair.
- Unknown name errors and lists declared names; imported STEP plus `--param`
  errors.
- Coercion for each supported type; a bad value errors.
- A `--param` run leaves the model's `__cadgen__` directory untouched.
- Params appear in the JSON output.

## Out of scope

- `--param` on `gen`/`refs`/`frame`/`measure`/`snapshot`. Persisted posed
  artifacts need parameters in the freshness key; `canonical_bake_hash` is the
  in-repo pattern when that day comes. Separate feature.
- Sweep syntax (`--sweep a=0:90:5`). A shell loop covers it.
- Detecting topology-changing params. Pose-only params keep occurrence refs
  stable across poses; a param that changes topology, say a hole count, may
  renumber refs. Document the limit rather than detect it.
- Sidecar (`.step.js` / `--params-path`) support. Sidecars are the render-time
  system: JS, viewer and snapshot only, invisible to the brep. No mechanism poses
  an imported STEP for interference today, and this feature does not add one.

Once this lands, the doc in this PR gets rewritten around the real flag (its
current `parameters.md` citation points at the sidecar system) and the mandate
becomes executable. Happy to review both halves.

earthtojake:
Hi @13thrule following up on this, let me know if you intend on completing the PR

PR #332, opened 2026-08-25

NgoQuocViet2001:
Re-files #327 against `develop`, per @earthtojake — that one targeted `main` and edited the generated copy under `skills/cad-viewer/`.

This one patches `packages/cadjs/src/lib/dxf/parseDxf.js` and adds the cases to the existing `parseDxf.test.js`.

## The bug

The paragraph-break replace carries the `i` flag:

```js
text = text.replace(/\\P/gi, "\n").replace(/\\~/g, " ");
```

`\P` is a paragraph **break**. Lowercase `\p` is paragraph **properties** — a different command that carries a payload up to a semicolon (`\pxqc;`, `\pxi-2,l2,t2;`). Because `p` is also absent from the inline-property class on the next line, the replace consumes only the two characters and leaves the rest in the text:

```
\pxqc;PART A        ->  "xqc;PART A"        (want "PART A")
\pxi-2,l2,t2;Item   ->  "xi-2,l2,t2;Item"   (want "Item")
```

Any MTEXT carrying a paragraph alignment or indent engraves the command payload. Centred titles (`\pxqc;`) and indented notes are the everyday cases.

## The fix

Make the break case-sensitive and add `p` to the property-run class:

```js
text = text.replace(/\\P/g, "\n").replace(/\\~/g, " ");
text = text.replace(/\\[fFhHcCtTqQwWaAp][^;]*;/g, "");
```

Order already works in favour of this: the `\P` replace runs first, so no uppercase `\P` survives to be eaten by the property run.

## Tests

Three cases appended to `parseDxf.test.js`, covering the properties command, the break, and the other property runs plus plain text.

```
after   ℹ tests 15   ℹ pass 15   ℹ fail 0
before  ℹ pass 14    ℹ fail 1    ✖ MTEXT paragraph properties are removed whole
```

Only the broken case fails on `develop` — the paragraph-break and other-property tests pass either way, which is what makes the case-sensitivity change safe.

The arc-sampling fix is re-filed as #331.


earthtojake:
Looks like one failing test case @NgoQuocViet2001 , can you fix that up or remove if the test case is no longer relevant?

earthtojake:
The code is right and the base is right now, thanks for the quick turnaround. One
step left: CI fails `bundle.sh --check` because `parseDxf.js` is baked into the cad
skill's esbuild bundle, and the source change makes the committed bundle stale. CI
names the fix:

```bash
scripts/bundle/bundle-skill.sh cad
```

Run that, commit the regenerated `skills/cad/scripts/packages/cadjs/bin`, and push.
Green after that and this merges. (#331 skips this only because
`buildDrawingLines.js` is not part of that bundle.)

NgoQuocViet2001:
Done — pushed `bf83636f` with the regenerated bundle.

Ran `scripts/bundle/bundle-skill.sh cad`; the only delta is in the minified `stripMtextFormatting` inside `skills/cad/scripts/packages/cadjs/bin/dxf-artifact.mjs`:

- `e=e.replace(/\\P/gi,"\n")` → `e=e.replace(/\\P/g,"\n")`
- `/\\[fFhHcCtTqQwWaA][^;]*;/g` → `/\\[fFhHcCtTqQwWaAp][^;]*;/g`

which is exactly the source change, so nothing else in the bundle moved. Pushed as a fast-forward on top of `faf30f3e` so the base and the reviewed commit are untouched.

One note in case it's useful to you: I first built the bundle against current `develop` (`cce04de6`, 0.4.28) and then against the PR head's own base, and `dxf-artifact.mjs` came out byte-identical (`sha256 9d7ffe2e…`), so the check isn't sensitive to the base drift here.

NgoQuocViet2001:
Correction to my last comment — regenerating `cad` alone was not enough, and CI was still red. Fixed in `7301e256`.

`parseDxf.js` is baked into **two** skill bundles, not one. After `bundle-skill.sh cad`, `bundle.sh --check` gets past the cad skill and then fails on the next one:

```
skills/dxf/scripts/packages/cadjs/bin is stale.
Run scripts/bundle/bundle-skill.sh dxf and commit skills/dxf/scripts/packages/cadjs.
```

I only saw this once I reproduced the CI job properly. My first pass ran `bundle-skill.sh cad` and the standalone cad check, which passes; `bundle.sh --check` walks every skill and needs the repo's own deps installed (`npm ci` in `packages/cadjs`, `packages/implicitjs`, `viewer`, `docs`) before it reaches the later ones. With that set up on Node 22 to match the workflow, the dxf failure showed up.

So the branch now carries two bundle commits, `cad` and `dxf`, with the same two-line delta in the minified `stripMtextFormatting`.

Verified locally at the pushed commit, clean tree:

```
$ scripts/bundle/bundle.sh --check
...
All bundle outputs are up to date.
$ echo $?
0
```

Both bundle commits sit on top of `faf30f3e` as fast-forwards, so the reviewed source commit is untouched.

NgoQuocViet2001:
Branch was showing `behind` — merged current `develop` (`cce04de6`, 0.4.28) in as `4f0020b6`, so you don't need to hit Update branch.

Worth re-checking rather than assuming, because `develop` moved `packages/cadjs/package.json` and its lockfile since this branch forked, and those pins feed the bundle. Re-ran the check on the merged tree with the repo's own deps installed:

```
$ scripts/bundle/bundle.sh --check
...
All bundle outputs are up to date.
$ echo $?
0
```

Both committed bundles are still correct against the new pins, so nothing needed regenerating a second time. PR diff is unchanged at 4 files — `parseDxf.js`, `parseDxf.test.js`, and the two `dxf-artifact.mjs` bundles.

earthtojake:
Could you retarget this at main? develop is retired and main is the only long-lived branch now. The tree also moved: packages/cadjs is packages/cadgen-js, and skills no longer vendor bundles, so the bundle regeneration you fought with is gone. The bug is still live at packages/cadgen-js/src/lib/dxf/parseDxf.js:681. The change is the same two regex edits plus your three cases in the parseDxf.test.js beside it. Green CI on main and I'll merge.

NgoQuocViet2001:
Refiled against `main` as #385 — closing this one.

You were right that the tree moved: `packages/cadjs` is `packages/cadgen-js`, and the two `skills/*/dxf-artifact.mjs` bundles this PR carried are gone. Retargeting the base would have produced a broken diff, so #385 is a fresh branch off `main` with the same two regex edits and the same three test cases.

One thing worth flagging, since it is the part that bit me last time: the bundle regeneration is not entirely gone. `parseDxf.js` is still vendored into two **committed** runtime bundles — `packages/cadgen/src/cadgen/_runtime/node/dxf-mesh.mjs` and `_runtime/browser/snapshot-render.js` — so `bundle.sh --check` does still gate this change. I regenerated both; the delta is exactly the two regex literals, and the check is green. It is only the `skills/` copies that no longer exist.

`node --test packages/cadgen-js/src/lib/dxf/parseDxf.test.js` → 15 passed, and reverting only the source fails the new test with `'xqc;PART A'`.

PR #333, opened 2026-08-25
viewer: stream asset and download responses in 64 KiB chunks

warun7:
## Summary

Split from #305 per review.

Asset and download responses (`GET /__cad/asset`, `/__cad/download`, and static dist files) read the whole file into memory before writing - a multi-GB GLB pinned that much RAM per request, under an unbounded `ThreadingHTTPServer`.

Responses now stream in 64 KiB chunks with an explicit `content-length` header, and a client that hangs up mid-download is absorbed (BrokenPipeError/ConnectionResetError) instead of raising from the handler thread. Nothing else about the response changes: still no-store, no etag/range, HEAD writes headers only.

The timeout work stays on #305 for the idle-watchdog rework discussed in review.

## Verification
- New `test_stream_file.py`: chunked payload integrity, disposition header, broken-pipe absorption
- Full `viewer/server_py` suite: 179 tests OK

### DIFF
--- a/viewer/server_py/server.py
+++ b/viewer/server_py/server.py
@@ -26,6 +26,7 @@ import argparse
 import os
 import posixpath
 import re
+import shutil
 import sys
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from urllib.parse import urlsplit, parse_qs, unquote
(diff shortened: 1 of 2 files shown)


earthtojake:
Verified live before merging: a 15.7 MB STL streams byte-identical in 239 chunks,
content-length matches disk, HEAD sends headers with an empty body, and the
download route carries the right disposition.

One nit for a follow-up rather than a blocker: `os.path.getsize` runs before
`open`, so a `replace_atomic` swap landing between the two mis-sizes
content-length. Opening first and sizing the handle with `os.fstat` closes the
race for free. Worth folding into your next pass on this file.

PR #335, opened 2026-08-25
skills: fix gcode's export remediation; add cad-viewer test suite; repair skill deps

warun7:
## Summary

Skill-level fixes and coverage gaps found in review.

### gcode remediation text pointed agents at a command that does not exist
The remediation emitted at runtime said `python scripts/step --kind part <input> --stl <output>.stl` - but the cad skill has no `scripts/step` and never had a `--kind` flag, so every STEP/DXF/URDF/SDF slicing flow got misdirected exactly when it needed help. The real CLI is `scripts/export <input> --stl <output>.stl` (gen sources and imported STEP/STP alike). Fixed in `gcode_tool.py`, `SKILL.md:123` and `references/slicer-backends.md`.

### cad-viewer was the only skill without a test suite
New `tests/python/skills/cad-viewer/test_packaged_viewer.py`:
- static pins on the packaged layout (vendored viewer runtime, npm `start` script + shim, Python launcher, requirements pinning the vendored cadgen) and on SKILL.md documenting the start command and port 3245
- a live smoke that runs `npm --prefix scripts/viewer run start` on a free port and asserts `/__cad/server` answers JSON with `backend=local-fs` on that exact port

### Skill dependency repairs
- `skills/implicit-cad/` shipped **no** requirements.txt although its documented snapshot CLI hard-imports playwright -> one added (`--editable ./scripts/packages/cadgen` + playwright), matching sdf/srdf/urdf.
- `skills/dxf/requirements.txt` was missing playwright for the same reason -> added.
- `snapshot_core.py`'s missing-playwright error told users to "Install the CAD skill requirements" - another skill's manifest. Now says "the invoking skill's own requirements.txt".

Dropped from scope since the scan: daemon `-` stdin inline handling is already documented in `cadgen_daemon/client.py`.

## Verification
- `scripts/test/test-python.sh`: all 15 suites OK, including the new **cad-viewer suite (5 tests)** wired into the runner via its `tests/python/skills/<skill>` directory
- gcode suite green against the rewritten remediation text

### DIFF
--- a/packages/cadgen/src/cadgen/snapshot_core.py
+++ b/packages/cadgen/src/cadgen/snapshot_core.py
@@ -772,7 +772,8 @@ class BatchSnapshotRenderer:
             except ImportError as exc:
                 raise SnapshotError(
                     "CAD snapshot requires the Python playwright package. "
-                    "Install the CAD skill requirements, then run `python -m playwright install chromium` if needed."
+                    "Install the invoking skill's own requirements.txt (it ships playwright), "
+                    "then run `python -m playwright install chromium` if needed."
                 ) from exc
             self.playwright = await async_playwright().start()
             self.browser = await self.playwright.chromium.launch(
--- a/skills/gcode/SKILL.md
+++ b/skills/gcode/SKILL.md
@@ -120,7 +120,7 @@ Pass `.stl`, `.obj`, and unsliced `.3mf` directly to the slicer. Convert `.ply`,
 
 Reject `.step`, `.stp`, `.dxf`, `.svg`, `.urdf`, and `.sdf` in v1. `inspect` and `slice` fail with a structured `remediation` object naming the skill and command that produce a sliceable mesh; use it instead of inferring a conversion workflow:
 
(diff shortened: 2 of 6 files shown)


earthtojake:
Looks like windows CI/CD is failing

earthtojake:
Merged, thanks for this and for chasing the Windows failure as far as you did.

The failure was in our launcher, not your test. `port_is_free` probed by connecting to the port and counted only `ConnectionRefusedError` as free. On the Windows runners a connect to a closed port fails in other ways (Hyper-V port exclusions, refusals arriving as timeouts), so every port read as occupied and the launcher refused to start. Your `_free_port` comment about ports binding fine and then reading as busy pointed straight at it: once your workarounds ruled out the ephemeral range and pre-bind verification, the probe was the only suspect left. #336 changes it to probe by binding, the same operation the server performs, and your smoke test passed on Windows with that fix underneath it.

warun7:
yeah that makes sense
i thought i was just running low on sleep lol
thanks

PR #338, opened 2026-08-26
release: enforce build-test rehearsal pairing; drop dead version/docs targets

warun7:
## Summary

Three small release-tooling fixes.

### Release workflow accepts invalid build-test dispatches
`AGENTS.md`/`CONTRIBUTING.md` say `target_branch=build-test` is strictly a CI/CD rehearsal that must be paired with `bump=none`, but nothing enforced it: dispatching with the default `bump=patch` burned a version number and opened a release PR into `develop`. The `release-pr` job now validates the pairing in its first step - before anything mutates - and fails with an actionable `::error::`.

### Dead version-sync targets
`sync-version.mjs` carried `required:false` entries under `skills/cad/scripts/packages/implicitjs/` - a mirror that exists nowhere (the cad skill ships only `cadgen` and `cadjs`). Removed. The `implicit-cad` pair stays: that path resolves through the vendored-runtime layout.

### Dead hero-GLB LFS branch in test-docs.sh
`.gitattributes` marks `docs/public/hero/**` deliberately **non-LFS**, so the script's `git lfs ls-files | grep docs/public/hero` gate never matched and the pull never ran. Removed with a comment so it stays gone.

## Verification
- `sync-version.mjs --check`: metadata synced from VERSION
- `scripts/bundle/bundle.sh --check`: all outputs up to date
- `scripts/test/test-global.sh`: 86 tests OK
- release.yml parses; guard step confirmed first after checkout
- test-docs.sh reaches its `npm run check` tail (local eslint gap only; CI covers)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,6 +75,20 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
+      - name: Validate build-test rehearsal pairing
+        env:
+          TARGET_BRANCH: ${{ inputs.target_branch }}
+          BUMP: ${{ inputs.bump }}
+          SET_VERSION: ${{ inputs.set_version }}
+        run: |
+          # AGENTS.md and CONTRIBUTING.md: target_branch=build-test exists to rehearse
+          # the CI/CD pipeline, so it must publish base_branch AS IT STANDS -- bump=none.
+          # Anything else burns a version number and opens a release PR for a rehearsal.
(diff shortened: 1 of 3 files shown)


earthtojake:
Closing. Since this opened, release.yml split into release-prepare.yml and release-publish.yml, a build-test dispatch now consumes a version on build-test by design, and the implicitjs entries left sync-version.mjs. The one piece still live is the hero LFS gate in scripts/test/test-docs.sh. Re-file that alone against main and I'll take it.

PR #339, opened 2026-08-26
moveit2_server: bound the warm MoveItPy instance cache

warun7:
## Summary

`MoveItPyAdapter` kept one warm `MoveItPy` instance per model **forever** - keyed by model asset hash, never evicted. Each instance is a ROS node plus a planning context, so serving many distinct robots crawled memory for the life of the server.

The cache is now bounded LRU:

- default limit **5** warm instances, tunable via `MOVEIT_PY_ADAPTER_CACHE` (floored at 1)
- a cache hit refreshes recency; the least-recently-used instance is shed when the bound is exceeded
- eviction calls the instance's `shutdown()` best-effort rather than letting the native ROS node be silently collected; a shutdown failure is swallowed so eviction always proceeds
- each model still pays its own first-touch cost - only the cross-model crawl is bounded

## Verification

- New `MoveItPyCacheTests`: LRU shedding order, recency refresh on hit, shutdown-failure tolerance (all without rclpy, matching the suite's existing style)
- Full MoveIt2 server suite: 20 tests OK; `scripts/test/test-python.sh` all suites green

### DIFF
--- a/tests/python/viewer/moveit2_server/test_moveit_py_adapter.py
+++ b/tests/python/viewer/moveit2_server/test_moveit_py_adapter.py
@@ -169,3 +169,72 @@ class MoveItPyAdapterHelperTests(unittest.TestCase):
 
 if __name__ == "__main__":
     unittest.main()
+
+
+class MoveItPyCacheTests(unittest.TestCase):
+    """The warm-instance cache is bounded LRU: heavyweight ROS nodes cannot be allowed
+    to accumulate one per model for the life of the server."""
+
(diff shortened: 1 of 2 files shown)


earthtojake:
@warun7 I've actually removed ros2 and moveit2 support on the release/0.5.0 branch, since i think this creates a bit too much bloat and probably isn't needed for the scope of the repo's skills. I'm ok to merge this in if you need it, but if not let's drop it.

warun7:
yeah if theres no support then it doesnt make sense
lets drop it

PR #343, opened 2026-08-30
fix(viewer): bind renders to canonical STEP bytes

amywork777:
## What changed

- Bind imported and generated Viewer packages to the current canonical STEP SHA-256.
- Keep generated packages detached from source-code freshness while still rejecting stale STEP bytes.
- Refuse stale generated packages on open instead of importing or rebuilding over the canonical artifact.
- Add coverage for missing and mismatched hashes.

## Why

The STEP is the source of truth. A render package may outlive the STEP it represents; showing that stale package can make the Viewer display the wrong geometry. Opening remains read-only and fails closed with the canonical cadgen rebuild guidance.

## Verification

- Viewer server suite: 55 passed, 1 skipped.
- Diff check passed.
- Reproduced with a fresh generated rat STEP in the Hardcore desktop app.


amywork777:
Superseded by 0.5.0: render trees are now keyed by the document bytes and schema. The current viewer test test_editing_the_file_unresolves_the_package covers the stale-STEP case this patch addressed, so the old descriptor-hash fix is no longer needed.

PR #348, opened 2026-09-03
fix(implicit): fall back to defaults for non-numeric graphics settings

NgoQuocViet2001:
## Problem

`clampNumber` calls `Number(value)` on whatever it is given, and `Number(null)`, `Number("")` and `Number([])` are all `0`. Since `0` is finite, the fallback never runs — the value clamps to the field's minimum instead:

| input | before | expected |
|---|---|---|
| `{ resolutionScale: null }` | `0.5` | `2` (default) |
| `{ resolutionScale: "" }` | `0.5` | `2` |
| `{ detail: null }` | `0.25` | `1.25` |
| `{ stepBudget: null }` (interaction) | `1` | `96` |

`undefined` and `NaN` already fell back correctly, which is what makes this easy to miss: the bug only shows up for `null` and `""` — exactly the two forms a persisted settings object produces. `undefined` round-trips through `JSON.stringify`/`parse` as either a dropped key or `null`, and an emptied numeric input in a settings UI arrives as `""`.

The effect is silent and backwards: a user who has never touched a graphics setting, but whose stored settings carry a null, renders at the *lowest* quality the limits allow. The step-budget case is the worst of them — a raymarching budget of 1 instead of 96.

## Fix

Take a value only when it is a number, or a string that holds one; anything else falls through to the default. Both `clampNumber` and the interaction step budget go through the same helper, since they had the same defect.

## Scope

`packages/implicitjs/src/lib/implicitCad/graphicsSettings.js` only. Real numbers, numeric strings, and genuine out-of-range clamping are unchanged — asserted in the new tests.

## Test plan

- Ran: `node scripts/run-tests.mjs src/lib/implicitCad/graphicsSettings.test.js` in `packages/implicitjs` → 4 passed.
- Checked: reverting only `graphicsSettings.js` fails both new tests (`expected defaults for null`), so they pin the reported behaviour.


earthtojake:
Closing. main no longer has packages/implicitjs or the implicit-cad skill, so graphicsSettings.js is gone and this has nothing left to patch. Your Number(null) analysis was right.

PR #350, opened 2026-09-04
feat: add robot component selection and references

realfishsam:
Named objects inside a robot's linked meshes now have a Components inventory with bidirectional viewport selection. The PR is rebased onto `main` and all viewer changes live under `apps/viewer`.

- Single-line rows grouped by robot link select and highlight individual mesh objects. Picking an object in the viewport selects and reveals its row; Ctrl/Cmd/Shift-click supports multiple selections.
- Objects retain their visual transforms while joints move. Repeated mesh instances receive distinct component identifiers.
- A Reference tab shows the selected components' link, visual, mesh, and object identifiers, with copyable references and an explicit clipboard failure message. Robot sheets support Components above and Reference/Joints below using the existing split-pane layout.
- Unnamed objects remain rendered but are filtered from the inventory. Components and Reference are omitted when nothing is named, including STL-only robots.
- Grouping uses memoized `Object.groupBy`, removing the repeated Map copies without mutating application data.

Robot references are self-describing prompt locators, for example `models/robot.urdf#link=arm&visual=arm%3Av1&object=3mf%3A0&index=0&name=bracket`. Their encoding and resolution are documented in the viewer README. They identify source mesh objects, not STEP topology, and are not accepted by the STEP selector CLI. Pose changes preserve references; reordering visuals or re-exporting meshes can change them.

Validation: viewer production build passed; `scripts/bundle/bundle.sh --check` passed; code review and whitespace checks completed. No tests were added or run, including the requested section-ordering case. Interactive browser behavior has not been manually verified.

### DIFF
--- a/apps/viewer/README.md
+++ b/apps/viewer/README.md
@@ -121,3 +121,28 @@ The backend's suite lives with cadgen and is not collected here; running only
 
 Headless UI verification uses Playwright with `--use-angle=metal` —
 the default software WebGL renderer is not what users see.
+
+## Robot components
+
+URDF, SRDF, and SDF files with named objects in their linked meshes expose a
+Components tab. Its single-line rows are grouped by link. Selecting a row
+highlights that mesh object in the viewport; picking it in the viewport selects
+and reveals its row. Ctrl/Cmd/Shift-click toggles additional components. Objects
+retain their visual transforms as joints move. Unnamed objects remain rendered
+but are omitted from the inventory; files without named objects have neither
+Components nor the component Reference tab.
+
+Reference shows the selected objects' link, visual, mesh, and object identifiers,
+with a copy button for each. Robot references are prompt locators of the form
+`models/robot.urdf#link=arm&visual=arm%3Av1&object=3mf%3A0&index=0&name=bracket`.
+The prefix identifies the robot file; path segments and fragment values are percent-encoded.
+`visual` is the parsed visual ID (`<link>:v<one-based visual ordinal>` for URDF);
+`object` is the mesh loader's object ID and `index` its zero-based position in the
+loaded mesh's parts list. `name` preserves the authored object name. Resolve the
+link and visual in the robot description to find the mesh file, then identify
+that mesh object using its name and index. Repeated mesh instances have distinct
+references because they belong to different visuals. SRDF locators name the SRDF
+file and resolve visuals through its paired URDF. SDF uses its parsed visual IDs.
+These locators describe mesh objects for prompts; they are not STEP face selectors
+and are not accepted by the STEP selector CLI. They remain stable across pose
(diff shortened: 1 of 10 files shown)


earthtojake:
Hey @realfishsam thanks for this! Did you build this tab for yourself and design it from scratch? Is it inspired by any other URDF visualization tools?

I would suggest modelling the design on the existing "Tree" component in the STEP file format, and also keeping each row in the tree to 1 line to reduce clutter. Note that STEP files also have a "Reference" tab (defaults to bottom-half) that may be helpful for showing more info about a selected component.

Thanks for adding this!

earthtojake:
Following up with review notes:

- Retarget at main. develop is retired and the viewer now lives in apps/viewer. The same four files exist there, so the port is mechanical.
- fileSheetSections.test.js covers the section ordering you changed. Add the components case there.
- STL meshes carry no part names, so on a robot with STL visuals the tab lists rows of "Unnamed component". Filter unnamed parts out, or skip the tab when nothing is named.
- The tab is read-only text right now. What would make it worth having is a selection and reference system for components like the STEP one: clicking a row selects and highlights the part in the 3D view, picking a part in the viewport highlights its row, and each component gets a copyable reference (STEP uses the form `bracket#o1.2.f6`) so you can paste it into a prompt and the agent knows which link and which mesh object you mean. The Reference tab I mentioned above is where the selected component's details belong.
- groupedComponents rebuilds the Map on every reduce step. A plain loop reads simpler.

Happy to review once it is against main.

realfishsam:
Thanks for the detailed review. Updated this in c67366c and retargeted the PR to `main`; the viewer changes now live under `apps/viewer`.

The Components tab now uses single-line rows grouped by link, with selection and highlighting in both directions between the list and viewport. Selected-component details and copy buttons live in Reference. Unnamed objects are filtered out, and the tabs are omitted when nothing is named. Grouping now uses memoized `Object.groupBy` instead of rebuilding a Map on each reduce step.

One design choice worth calling out: the copied references explicitly identify the robot file, link, visual, and mesh object. They're documented prompt locators, not STEP selectors accepted by the existing selector CLI.

The production build and bundle freshness check passed. I haven't added the requested section-ordering test or run tests, and the browser interactions still need manual verification. Happy to hear whether this reference format fits what you had in mind.

earthtojake:
Mind sending some screenshots? I'll run this locally too over the weekend and give a more detailed review!

earthtojake:
We've taken this over the line in #402: your commit is kept as-is, plus the rebase onto current main, the tab-layout test fixes, the section-ordering test, a Render-mode gate so Materials keeps per-visual targets, and a non-throwing path for a bad mesh part. You're credited as co-author on the follow-up commits. This PR will close when #402 merges; thanks for building it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #370, opened 2026-09-07
Add tendon hand workbench and continuous tube rendering

earthtojake:
> **Superseded review:** the final CAD runtime, caching and viewer changes are in [PR #388](https://github.com/earthtojake/text-to-cad/pull/388). The Tendon Hand project landed separately in [PR #384](https://github.com/earthtojake/text-to-cad/pull/384). This older branch is preserved for history; its description and validation below reflect an earlier state.

Adds a 24-DOF, 48-tendon hand modeling workbench and the runtime support needed to render continuous deformable tendons. The runtime changes add CPU/GPU tube deformation and braided appearance, reuse assembly mesh buffers, bound snapshot socket writes, and preserve the active parameter domain when exporting periodic NURBS surfaces.

Includes the model sources, authored presentation configuration, validation runners, Markdown evidence tables, and `BUGS.md` through BUG036. Generated STEP/BREP files, sidecar JSON, numeric validation dumps, route packets, caches, and logs are gitignored. This branch adds no LFS payloads.

### Readiness

This is a **draft**. Local R13 checks cover 3,259 solid occurrences and 225 specified static poses; the recorded export QA does not claim exact global source/export material equality. Independent visual acceptance, final pose/explode sequences, and portable regeneration remain unfinished. The reconstruction currently depends on local intermediate exports and metadata.

**BUG036 remains unresolved:** the full hand preview expands to 19.3 million triangles and 2.50 GB of CPU render arrays before other allocations, and the user reported a browser crash. Bounded initial loading must be implemented and validated outside the user's browser before another interactive handoff. The earlier buffer-sharing fixes do not resolve this remaining limit.

### Validation

- `scripts/test/test.sh` — passes: 806 shared-JS tests, 347 Viewer client tests, and all Python/package/skill/global suites (one skipped global test).
- `scripts/bundle/bundle.sh --check` — passes.
- `scripts/release/check-version.sh` — passes; VERSION remains 0.5.0.
- `scripts/github-workflows/check-builds.sh` — passes.
- `npm --prefix apps/viewer run build` — passes (existing chunk-size warnings).
- All 504 hand Python source files parse; every input hash in the local current static certificate was verified unchanged.

Start review with `models/assemblies/src/anthropomorphic_hand/README.md`, `PR_CONTENTS.md`, and `BUGS.md`.


earthtojake:
Review session handoff: see `HANDOFF-viewer-perf.md` at the repo root of this branch (delete before merge). Summary:

- BUGS.md triaged (verdict table in the notes repo); cadgen defects fixed and pushed: warm `read_step` served build shapes (028/029), op_memo broke `is_same` across ops (011), snapshot uploads crossed the Playwright pipe (006/018), pole-hull bounds (004), validate under mirror/scale (021), GLB Map cap (015), plus docs/CLI records.
- Viewer memory work landed (indexed geometry, line-pass edges, progressive publish, component-keyed GPU cache): hand first geometry 86 s to 4.7 s, renderer peak 4.5 to 3.4 GB, GPU 2.4 GB to 0.5 GB.
- OPEN: the user's Chrome still crashes loading the hand (needs load-time peak well under 2 GB); the final-publish render-module gating in `52c4ad771` must be reversed (dead animation controls / kinematics tab bounce while loading); full-load time is 2.1x baseline; orbit at ~10k draw calls. Unverified WIP for incremental records + instanced edges is on `wip/pr370-viewer-perf-instanced-edges`.

earthtojake:
Second review session. Everything below is pushed to this branch; `HANDOFF-viewer-perf.md` is rewritten around what is measured (delete it before merge).

## Viewer memory and load, measured on the r13 hand (866 components / 3,259 occurrences)

| metric | baseline (old client) | at `fb1632c5b` | now |
|---|---|---|---|
| first geometry on screen | ~86 s | 4.7 s | **4.9–5.9 s** |
| time to fully loaded | 86–90 s | 186 s | **147 s** |
| renderer RSS peak | 4.5 GB | 3.4–3.6 GB | **2.65 GB** (largest single process 2.58 GB) |
| renderer RSS after a forced GC | — | — | **2.22 GB** (heap 1.18 GB) |
| JS heap peak | 3.3 GB | 2.2–2.4 GB | **1.72 GB** |
| GPU buffer bytes | 2418 MiB | 524 MiB | **507 MiB** |
| publishes | 1 | 28 | **8** |
| draw calls over the load | — | 368k | **98k** |

It also loads in a real browser without crashing (144 s, no console errors). **The previous session's "well under 2 GB" gate is still not met** — 2.65 GB peak, 25 % below the 3.6 GB that crashed Chrome twice, but not under the bar.

What landed:

- **The tube braid shader never compiled.** `bbc48ff9c` lifted the shared varying's declaration into the stage hook, which anchors on the same `#include <common>` the braid block uses — so the fragment stage read `vCadTubeMaterial` above the line declaring it and *every braided tube lost its program*. The hook now declares last; a test pins declaration-before-use.
- **Publish batches double** (8 components / 8 MB up to 256 / 128 MB). A publish recomposes the package and walks every occurrence already on screen, so a fixed 32 made the publish *count* grow with the model too. The count is now logarithmic and the first batch is *smaller* than the old fixed one, so first geometry arrives sooner.
- **The surf worker pool is released** when a load ends or aborts, instead of eight isolates holding the heap each grew for its largest component.
- **A deformed tube's edges honour `edges.classes[*].thickness`** — one screen-space fat line per drawn class instead of one 1-px vertex-coloured GL_LINES. On this model the tubes are the tendons, so they were the one thing ignoring the thickness control.
- **The camera frames again once every component has arrived**, unless you have already moved it (the hand used to settle at 191 % with most of itself off-screen).
- Stale BVH queue entries, a quadratic free-slot walk on mass release, and the memory probe now accounting for picking.

Two diagnostics worth more than the fixes:

- **The load is worker-CPU-bound, not transport.** The `/__tess_cache/` route serves 315 MiB/s to one reader and 681 MiB/s across eight, while a load only draws ~15 MiB/s. And the cache earns its keep — blocking it takes the routing fixture from 32 s to 63 s. The cost is per-component: entry decode, `buildMeshDataFromSurf`, `buildSelectorBundleFromSurf`.
- **Picking is not the memory gap.** With `faceIdBytes`/`pickBytes` in the probe, both are *zero* after a full load: selector topology is lazy. That was the largest named suspect for the ~1.4 GB the accounting could not see. What is left points at the per-occurrence object graph — 3,259 meshes, 5,061 materials — which is also what the remaining draw calls are.

## The hand animates

The render module opened with "Static surface presentation only; no joint choreography". It now carries five motions and a tour of all of them: fist, finger roll, thumb-to-index pinch, splay-and-gather, wrist flex.

The motion is exact. The module carries the build's own authored joint datums and evaluates `lib.layout.assembled_transforms` in JavaScript; the artifact is written at pose zero, so a frame's transform at a pose *is* its delta from rest. Generated by `write_showcase_presentation.py` — edit the generator, never the module.

**The tendons do not bend, and the module says so.** A posed route is a routing solve, not a transform: moving each route group rigidly with its own frame tears the centerline apart by up to 100 mm (a group spans from the *previous* guide to its own), and `deformTube` rightly refuses a centerline that is not tangent-continuous to 1e-5 mm. So the cords are shown where they are true and faded where they would lie, with the pair driving each joint staying brightest. Re-solving all 48 routes per keyframe is the follow-up that would let them bend.

## `cadgen step snapshot --video`

`--animation <clip> --time <s>` rendered one frame; the kinematics reference ended with "there is no sequence output". Now:

```bash
cadgen step snapshot STEP/arm.step demo.mp4 --animation demo --video '{"fps": 60, "quality": "high"}'
cadgen step snapshot STEP/arm.step demo.gif --animation demo --video '{"fps": 12, "seconds": 3, "start": 1.5, "quality": "draft"}'
```

The OUT extension picks the container; keys are closed (`fps`, `seconds`, `start`, `quality`, `loop`); ffmpeg is checked *before* the first frame. The source and model are prepared once for the whole sequence, and the camera is fitted once to the union of every frame's bounds and held — a still's per-pose auto-fit is exactly the breathing a video must not have. The hand's 30 s showcase renders 1,800 frames at 1080p60 in about five minutes.

Three independent reviews found eleven defects in the first cut (a `.mp4` OUT accepted for a still and written as a PNG; `seconds` defaulting past the end of a clip when `start` was set; the render silent for minutes); all are fixed.

## Suites

`packages/cadgen-js` 837, `apps/viewer` 347, Python 1723, global 125, `scripts/bundle/bundle.sh --check` clean.

PR #385, opened 2026-09-11

NgoQuocViet2001:
Refiles #332 against `main` as you asked. Same fix, rebuilt on the current tree.

## Problem

`stripMtextFormatting` matches the paragraph break case-insensitively:

```js
text = text.replace(/\\P/gi, "\n")
```

`\P` is a paragraph BREAK. Lowercase `\p` is paragraph PROPERTIES — a different command that carries a payload up to a semicolon (`\pxqc;`, `\pxi-2,l2,t2;`). The `i` flag makes the break pattern match it, consuming only those two characters and leaving the payload in the text:

| MTEXT source | engraved before | expected |
|---|---|---|
| `\pxqc;PART A` | `xqc;PART A` | `PART A` |
| `\pxi-2,l2,t2;Item` | `xi-2,l2,t2;Item` | `Item` |

Paragraph properties are what AutoCAD writes for centred or indented MTEXT, so any drawing whose text is centred engraves the alignment code as part of the label.

## Fix

Two edits in `packages/cadgen-js/src/lib/dxf/parseDxf.js:681-686`:

- drop the `i` flag so `\P` only matches the break
- add `p` to the inline-property-run class, so `\p...;` is removed whole the way `\f...;` / `\H...;` already are

## Scope

One function plus its tests, and the two committed runtime bundles that vendor it (`_runtime/node/dxf-mesh.mjs`, `_runtime/browser/snapshot-render.js`). The bundle delta is those same two regex literals — `bundle.sh --check` is green.

## Test plan

- Ran: `node --test packages/cadgen-js/src/lib/dxf/parseDxf.test.js` → 15 passed (12 pre-existing + 3 new).
- Checked: reverting only the source fails the first new test with `'xqc;PART A' !== 'PART A'`.
- Ran: `scripts/bundle/bundle.sh` then `scripts/bundle/bundle.sh --check` → `All bundle outputs are up to date.`
- The three new cases cover paragraph properties removed whole, `\P` still producing a newline, and `\H...;` / `\C...;` / plain text unchanged.

Closes #332.


earthtojake:
Verified against current main: the regex change is right and all 15 DXF tests pass with it. The two bundle files under `packages/cadgen/src/cadgen/_runtime` conflict with main now, and we're about to stop tracking those bundles on main. Hold off on rebasing until that lands; I'll ping here, and a rebase that drops the bundle files from the diff is all it needs to merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

earthtojake:
Rebased for you as #411 since the branch couldn't be updated from this side and the two `_runtime` bundles it patched no longer exist on `main` (#399 made them build outputs). Your commit and authorship are kept as-is, minus the bundle edits. This PR closes once #411 merges.

---
_Generated by [Claude Code](https://claude.ai/code)_

earthtojake:
Merged as 366937e via #411, with you co-credited on the commit. Thanks for the fix and for the clear write-up of why the `i` flag was the bug — the before/after table made it easy to verify.

This will ship in the next release; 0.6.2 went out just before the merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #390, opened 2026-09-15
Fix the tom-cad FEEDBACK.md bugs (2026-09-15 audit)

earthtojake:
Verifies and fixes the bugs reported in `~/robots/tom-cad/FEEDBACK.md` (audited against cadgen 0.5.1). Each was reproduced on an isolated fixture before the fix; every fix carries a regression test that fails on the pre-fix code.

## Confirmed bugs, fixed

| # | Report | Root cause | Fix |
|---|---|---|---|
| 1 | Nested kinematic targets receive upstream motion twice | `kinematicsModule.update` applied every accumulated world delta to every part under its target; a part inside two nested targets composed two absolute deltas | Each part takes the delta of the most specific mate naming it (`kinematicsDeltaTargets`), resolved through the effects api's own target resolution. tom-cad's `build_kinematics` workaround can be dropped once this ships. |
| 2 | Arc measurement mixes local and world coordinates | `assembly_lookup` placed only `origin` params; analytic types spelling their point `center` (circle, sphere, torus, ...) stayed component-local and `measure` consumed them | Every point/direction key the geometry writer emits is placed |
| 4 | Deferred model children are empty before geometry resolution | `LazyCompound` forced only on `_wrapped`; anytree serves `children`/`descendants`/`leaves`/`is_leaf`/`height`/`size`/`repr` from node state that `_force` fills | Force at `__getattribute__` against a closed deferral vocabulary; placement/label/color assignment still defer |
| 5 | Group labels resolve in kinematics but fail in STEP refs | Label aliases were built from leaf occurrence rows only | Interior nodes are indexed through the same funnel; `_group_id` canonicalizes a label first |
| 6 | Snapshot JSON guidance omits selection nesting | Doc gap + generic "unknown key" error | Error names the `selection` nesting; `snapshot-review.md` shows the JSON form |
| 7 | Robot snapshot renders fragmented link geometry, camera too zoomed in | `buildUrdfMeshGeometry` never declared `partTransformsBaked: false`, so the renderer drew every link unscaled at the origin while the camera framed the correct posed bounds | Flag set on both branches; `_runtime/browser` regenerated. Same function backs the viewer's URDF and SRDF views. |

## Historical reports

- **DXF closure splits coincident endpoints at a rounding tie**: confirmed and fixed (`drawing_checks.py` merged endpoints by exact rounded-key equality; now integer-bucket adjacency at the existing tolerance).
- **Early-kernel hint blames cadgen internals**: confirmed and fixed (`kernel_import_site.py` now skips every frame under the cadgen package).
- **Snapshot daemon hides worker failure**: already fixed by `9235973cb` on `codex/tendon-hand-performance`; not on `main` yet.
- **build123d tessellation assumes triangulation**: cadgen never calls `Shape.tessellate`; dependency-owned.
- **Playwright string limit**: bulk assets already go over the loopback asset server; only the package descriptor still crosses `page.evaluate`, which scales with occurrence count, not mesh size. Not reproduced.

## Not a cadgen bug

- **#3 `Compound.intersect` ignores a moved parent** is build123d issue #1394, fixed upstream in `217db07e` but not in any tagged release (we pin 0.11.1). cadgen's `inspect interfere` runs its own OCP booleans on placed shapes and is not exposed. Documented as a gotcha in `build123d-modeling.md`.

## Also

- `scripts/test/test-global.sh` now puts the checkout's `packages/cadgen/src` on the path like `test-python.sh` does; from a worktree it was testing the primary checkout's cadgen through the venv's editable install.

## Verification

- `scripts/test/test-python.sh`: 1236 cadgen package tests + every skill suite OK
- `npm --prefix packages/cadgen-js test`: 784 pass; `npm --prefix apps/viewer run test`: 347 pass
- `scripts/test/test-global.sh`: 125 OK; `scripts/bundle/bundle.sh --check`: up to date
- `cadgen urdf snapshot tom.urdf` on tom-cad renders the full arm correctly framed; the gripper-horn reproducer reports 0 mm error on the original (un-worked-around) sidecar.


earthtojake:
Superseded: these commits were merged into the PR #388 branch codex/tendon-hand-performance as 031a421b2 (plus the viewer joint-posing fix and browser gate), per Jake's direction that the tom-cad feedback work lands on the perf branch.

PR #391, opened 2026-09-16
One DFM skill for sheet metal, CNC and injection molding, with a measurement tool that holds up

amywork777:
## What this is

One skill, `dfm`, that reviews a part for **sheet metal**, **CNC machining** or **injection molding** — with measured evidence, the cited rule behind every finding, and an explicit **unverified** for anything it could not measure. Plus `mold_tool.py`, a fact-only geometry tool for the molding half.

## Since your review

**One skill.** The three `references/process-review.md` files were 88% identical and "Evidence first" and "Report and redesign" repeated across three `SKILL.md` files. Now:

```
skills/dfm/
  SKILL.md                            # evidence rules, rule selection, report format, which reference to read
  references/sheet-metal.md           # that process's checklist, its two sources, its worked example
  references/cnc.md
  references/injection-molding.md     # plus how to read every fact the tool reports
  scripts/mold_tool.py
  requirements.txt                    # SKILL.md says only the molding measurements need it
  agents/openai.yaml
```

All three processes are named in the description, so it still triggers on "can this be bent", "can this be machined" and "can this be molded". A fourth process costs one reference file. Tests moved to `tests/python/skills/dfm/`, the README collapsed to one row, the hand-written docs list gained it, and it joined the cadgen-free list in `AGENTS.md`.

`dfam-check` is untouched here, as you said — the name `dfm` is fixed now so nothing gets renamed twice.

`models/dfm-process-comparison/reports/` is deleted, along with the README's two references to `cadgen step inspect validate`. The model, `scripts/verify.py` and the headline numbers stay.

## `mold_tool.py`, against your battery

Every row below is now a test whose truth comes from the construction.

| Finding | Fix | Measured |
| --- | --- | --- |
| a curved face tangent to the pull reported mesh-dependent zero-draft area | a tangent band is separated from a flat zero-draft wall: at a tangent line the surface passes THROUGH parallel, so the face has smoothly-joined neighbours leaning both ways | sphere 0.00 mm² of wall at both resolutions (tangent band reported apart, and it halves as the mesh refines); a cylinder's wall still counts, 1884.2 against 1884.96 |
| `min_wall_draft` read one facet | pooled by normal over the bulk of the wall area, above an area floor; the single worst triangle stays as `min_facet_draft` | a 2° box with R3 top fillets reads 2.0, not the fillet's 1.3 sliver |
| a part modelled off-axis had no good pull among x, y, z | `pulls` adds the mesh's own principal axes | a 2° part tilted 20° reads 2.0 on a principal axis |
| wall area by opening direction was per-face only | `wall_area_by_opening_mm2` (`+pull` / `-pull` / `none`) | — |
| one ray per triangle counted whole triangles | faces are split to a small maximum edge before casting, rays still cast against the original surface, with a face budget | shelf over a plate **7988** against a true 8000 (was 9500); dovetail floor strips **586** against 582 (was 402) |
| a two-body STL reported 2400 mm² of undercut | each connected body is tested against itself; `body_count`, `per_body` and a note say so | two plain plates: **0** |
| the 0.25 mm raster printed 0.0 in metres and took 2.2 GB for a 2 m panel | raster sized to the part and capped | metres: 8000 mm² correct; 2 m panel: 3,000,000 mm² in 2.5 s |
| wall thickness failed in the opposite direction from `dfam_tool` | the family is **removed**. `_wall_facts`, `_cone_directions` and the thickness paragraph are gone; the reference points at `$dfam-check`'s `measure` with a note to ignore its `max_mm` | — |
| a degenerate triangle put bare `NaN` in the JSON | dropped on load, and JSON is written with `allow_nan=False` | a strict parser accepts the report |
| a dead family exited 0 | `"partial": true`, `partial_sections`, exit 2 | — |
| `--pull -z` was read as a flag; `--pull ""` raised `IndexError` | argv normalised before argparse; empty and sign-only pulls report | — |
| a 4x3x2 inch part passed `units_suspect` | `--units mm\|cm\|m\|in\|ft`, echoed as `scale.declared_units` | — |
| "a fine tessellation" | a face budget, 50k–200k triangles | — |

Runtimes on the fixtures: 1–4 s per `measure`.

## One more, from running it on the repo's own example parts

Measured against construction: the circular flange's projected area 4143 mm² against 4150 computed, the L-bracket's undercut 317.9 mm² against the 301 mm² of its two back-plate bores, the enclosure's 6997 against 6996.6.

But the flange's **outer rim — 1759 mm² of real zero-draft wall — was filed as a tessellation artifact.** Its R1.5 fillets join it tangentially top and bottom, so its facets have smoothly-joined neighbours leaning both ways, which was the whole test for a tangent band. Wrong in the dangerous direction: draft is exactly what that rim needs.

Size settles it. A band standing in for a tangent line is a facet of the same curved surface as the facets bounding it, so it reaches about as far along the pull as they do; a wall is much taller than the fillet that ends it. Measured: **1.00 to 1.24** for three tessellations of a sphere, **47.6** for that rim — the cut at 3 sits in a gap of nearly forty.

Two candidates rejected on the way, each with the measurement that rejected it: the band's height against the part's depth (coarse sphere 17%, rim 70% — too close), and its height against its own width across the pull (a box's wall reads 0.5 to 0.67, *below* the spheres' 1.03 to 1.45; a wall can be wide and short).

Flange now reads 3831 mm² = bore 942 + six bolt bores 1131 + rim 1759, each matching its construction.

## Checks

`tests/python/skills/dfm/test_mold_tool.py` (31 tests) and the `tests/python/global` policy gates pass locally.


earthtojake:
Thanks Amy. CI is green after the main merge. I read the three skills, test-merged the branch, and ran `mold_tool.py` against 41 sample parts whose answers I know from construction, plus your tray at two mesh resolutions.

The skills hold the right line: no hardcoded thresholds, the shop's spec wins, and anything unmeasured reads **unverified**. The tool's draft, undercut and projection families are accurate. Wall thickness should leave this PR. Four changes to the PR's shape first.

## Shape

**1. One skill, named `dfm`.** The three `references/process-review.md` files are 88% identical, and "Evidence first" and "Report and redesign" repeat across the three `SKILL.md` files. Merge them:

```
skills/dfm/
  SKILL.md                      # shared evidence rules, report format, which reference to read
  references/sheet-metal.md     # the process-specific bullets and the two source links
  references/cnc.md
  references/injection-molding.md
  scripts/mold_tool.py
  requirements.txt              # say in SKILL.md that only the molding measurements need it
  agents/openai.yaml
```

Name all three processes in the description so the skill still triggers on "can this be bent", "can this be machined" and "can this be molded". A fourth process then costs one reference file. Move the tests to `tests/python/skills/dfm/`, collapse the README rows to one, add the skill to `apps/docs/src/app/page.tsx` (that list is hand-written), and add it to the cadgen-free list in `AGENTS.md`.

**2. `dfam-check` folds into `dfm` in a follow-up PR, which we will drive.** `mold_tool.py` started as a copy of `dfam_tool.py`: `_load` is identical, `_mesh_facts`, `_scale_hint` and `_safe` are near-copies, and the two `requirements.txt` files match. Skills cannot share code, so the copies drift. The follow-up adds `references/additive.md` and one tool with a shared core. You do not need to touch `dfam-check` here. The name `dfm` is chosen now so nothing gets renamed twice.

**3. Protolabs and Hubs as the fallback sources are fine.** Keep them.

**4. Delete `models/dfm-process-comparison/reports/`.** Keep the model, `scripts/verify.py` and the README's headline numbers. The five `*-validation.json` files came from `cadgen step inspect validate`, which #425 removed, so drop the README's two references to it as well. `verify.py` already proves each STEP reimports as one valid solid.

## `mold_tool.py`: results from the 41 parts

**Draft: exact on flat and conical faces.** Boxes at 0, 0.03, 0.5 and 2 degrees, a cylinder, a 3 degree cone, a filleted drafted box and a vector pull on a rotated part all matched. Four gaps:
- A curved face tangent to the pull reports zero-draft area that depends on the mesh. One sphere read 261 mm² coarse and 67 mm² fine. Separate "tangent line on a curved face" from a flat zero-draft wall.
- `min_wall_draft` reads one facet. Your injection tray read 0.97 degrees on one mesh and 0.60 on another, from 0.05 mm² facets. Report a minimum over pooled faces above an area floor.
- A 2 degree part modelled 20 degrees off-axis reads 12.3 under `--pull z`, and `pulls` tries x, y and z with no hint that none fit. Add the mesh's principal axes to `pulls`.
- Add wall area by opening direction (`+pull` / `-pull`). Today it appears per face only.

**Undercuts: the logic is right.** Through and blind holes along the pull read 0, a snap hook with a shut-off window reads 0, a side hole reads 37.67 against 37.70, and a groove, an hourglass and a sealed void land within 1%. Two gaps:
- One ray per triangle centre counts whole triangles. On CAD-coarse meshes a snap hook read 110 against 140, a dovetail 402 against 582 (it missed the floor strips), and a shelf over a plate 9,500 against 8,000. The shelf subdivided four times read 8,094. Subdivide to a maximum edge length before casting, or sample several points per triangle.
- A two-body STL of two plain plates reports 2,400 mm² of undercut. Warn or refuse when `body_count > 1`.

**Projection: within 1% on all four cases,** including a diagonal pull and a 3 mm pin. The raster is a fixed 0.25 mm, so a part exported in metres prints area 0.0 and a 2 m panel takes 2.2 GB for a 12-triangle box. Scale the raster to the part and cap the cells.

**Wall thickness: remove the family from this PR.** I ran the same parts through `dfam_tool.py`, which casts one ray along the normal:

| Part (truth) | `dfam_tool` | `mold_tool` |
|---|---|---|
| 3 mm plate | min 3.0, p05 3.0, max 60 | min 0.30, p05 2.07, max 3.0 |
| Solid 60 mm cube | 60 / 60 / 60 | p05 3.0, median 20.7 |
| Wedge, 1 to 5 mm | min 1.17, max 60 | min 0.10, max 5.4 |
| 1.5 mm rib on a 2.5 mm plate | min 1.5, max 60 | min 1.18, max 7.1 |
| Your injection tray (thinnest rib about 1.18) | min 1.11 | min 0.97, 0.61, 0.30 across three meshes |
| Cube with a sealed void, 10 mm wall | 40 | max 40 |
| Peak memory | about 110 MB | 0.6 to 4 GB, 43 GB at 82k faces |

The two methods fail in opposite directions. The single ray gets the thin end right and reads a plate's length as `max`. Your cone fixes `max` and breaks the thin end, because tilted rays leave through a neighbouring face, and `SKILL.md` tells the agent to quote `p05_mm`. trimesh's numpy ray engine also allocates rays x faces, and the cone multiplies the rays by nine. Neither tool sees a void, since `mesh.split()` casts each body against itself. The follow-up will take the thin end from the single ray and the thick tail from the cone, cast against the whole mesh, in batches. For this PR, delete `_wall_facts` and `_cone_directions`, delete the thickness paragraph in `SKILL.md`, and point the agent at `$dfam-check`'s `measure` for wall thickness when that skill is installed, with a note to ignore its `max_mm`.

**Robustness.**
- A plain build123d STL of a part with a sphere carries 2 zero-area triangles, and the report then holds bare `NaN` tokens that strict JSON parsers reject. Drop degenerate faces on load and write JSON with `allow_nan=False`.
- `dfam_tool.py` marks a report `"partial": true` and exits 2 when a family errors. The copy lost that, so `mold_tool.py` exits 0 with a dead family. Copy `_mark_partial` back.
- `--pull -z` fails because argparse reads `-z` as a flag, and only `--pull=-z` works. `--pull ""` raises `IndexError`.
- A 4 x 3 x 2 inch part passes `units_suspect`. The tool cannot infer inches, so take a `--units` argument or state the millimetre assumption in the report.
- `SKILL.md` asks for an STL "at a fine tessellation". Give a face budget in its place.

**Tests.** The 15 current tests pass because none probes these cases. Add the failing fixtures as tiny test-owned parts: coarse against subdivided undercut, the dovetail floor, two bodies, a degenerate triangle, the off-axis part, the sphere's zero-draft area, the `--pull` forms and the metres projection. My harness is below. Each fixture states its truth and prints the tool's answer beside it. The `t_` fixtures cover thickness and belong to the follow-up.

<details>
<summary>battery.py (41 fixtures, build123d + trimesh, runs the CLI)</summary>

```python
"""Edge-case battery for PR #391's mold_tool.py: fixtures with known answers, tool run as a CLI.

Each fixture states the TRUTH (from construction) for the family it targets; the harness prints
truth vs. what the tool reported and a verdict. Run: python battery.py [name-prefix ...]
"""
from __future__ import annotations

import json
import math
import re
import subprocess
import sys
from pathlib import Path

import numpy as np
import trimesh
from build123d import (Axis, Box, Cone, Cylinder, Plane, Polygon, Pos, Rectangle, Rot, Sphere, export_stl, extrude,
                       fillet, revolve)

HERE = Path(__file__).parent
FIX = HERE / "fixtures"
FIX.mkdir(exist_ok=True)
PY = sys.executable
TOOL = HERE / "mold_tool.py"


def stl(shape, name, tol=0.08, ang=0.3):
    path = FIX / f"{name}.stl"
    export_stl(shape, str(path), tolerance=tol, angular_tolerance=ang)
    return path


def tm(mesh, name):
    path = FIX / f"{name}.stl"
    mesh.export(path)
    return path


def drafted_box(l=40, w=30, h=20, deg=2.0):
    return extrude(Rectangle(l, w), h, taper=deg)


def cup():
    return drafted_box() - extrude(Rectangle(36, 26), 18, taper=2.0)  # open at z=0, 2 mm walls and roof


def hook(window=False):
    part = Pos(0, 0, 1) * Box(40, 20, 2) + Pos(0, 0, 7) * Box(3, 10, 10) + Pos(3, 0, 11) * Box(3, 10, 2)
    if window:
        part -= Pos(3.1, 0, 1) * Box(3.2, 10.4, 2)  # flush with the post wall at x=1.5
    return part


def shadow():
    return Pos(0, 0, 2.5) * Box(100, 100, 5) + Pos(-45, 0, 15) * Box(10, 100, 20) + Pos(-30, 0, 27.5) * Box(40, 100, 5)


def get(report, dotted):
    node = report
    for key in dotted.split("."):
        if not isinstance(node, dict) or key not in node:
            return None
        node = node[key]
    return node


FIXTURES = []


def fixture(name, family, pull="z", samples=300, extra=()):
    def wrap(fn):
        FIXTURES.append({"name": name, "family": family, "pull": pull, "samples": samples, "extra": list(extra), "build": fn})
        return fn
    return wrap


# ---------------------------------------------------------------- DRAFT
@fixture("d_box_0deg", "draft")
def _():
    return stl(Pos(0, 0, 10) * Box(40, 30, 20), "d_box_0deg"), [("draft.zero_draft_wall_area_mm2", 2800, 0.01)], "4 vertical walls"


@fixture("d_box_2deg", "draft")
def _():
    return stl(drafted_box(deg=2), "d_box_2deg"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.drafted_wall_mean_draft_deg", 2.0, 0.01)], "2 deg taper"


@fixture("d_box_0.5deg", "draft")
def _():
    return stl(drafted_box(deg=0.5), "d_box_05deg"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.drafted_wall_mean_draft_deg", 0.5, 0.02)], "0.5 deg taper"


@fixture("d_box_0.03deg", "draft")
def _():
    return stl(drafted_box(deg=0.03), "d_box_003deg"), [("draft.zero_draft_wall_area_mm2", 2800, 0.02)], "0.03 deg < zero_tol 0.05: should pool as zero draft"


@fixture("d_cylinder", "draft")
def _():
    return stl(Pos(0, 0, 15) * Cylinder(10, 30), "d_cylinder"), [("draft.zero_draft_wall_area_mm2", 2 * math.pi * 10 * 30, 0.02)], "upright cylinder wall"


@fixture("d_cone_3deg", "draft")
def _():
    top = 15 - 30 * math.tan(math.radians(3))
    return stl(Pos(0, 0, 15) * Cone(15, top, 30), "d_cone_3deg"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.drafted_wall_mean_draft_deg", 3.0, 0.02)], "3 deg cone"


@fixture("d_sphere_coarse", "draft")
def _():
    return tm(trimesh.creation.icosphere(subdivisions=2, radius=20), "d_sphere_coarse"), [("draft.zero_draft_wall_area_mm2", 0, 0)], "sphere: zero-draft is a LINE (area 0); any area is tessellation noise"


@fixture("d_sphere_fine", "draft")
def _():
    return tm(trimesh.creation.icosphere(subdivisions=4, radius=20), "d_sphere_fine"), [("draft.zero_draft_wall_area_mm2", 0, 0)], "same sphere, finer mesh"


@fixture("d_uvsphere", "draft")
def _():
    return tm(trimesh.creation.uv_sphere(radius=20, count=[32, 32]), "d_uvsphere"), [("draft.zero_draft_wall_area_mm2", 0, 0)], "uv sphere: an equator band of facets is exactly vertical"


@fixture("d_filleted_2deg", "draft")
def _():
    box = drafted_box(deg=2)
    top_edges = box.edges().filter_by_position(Axis.Z, 19.9, 20.1)
    return stl(fillet(top_edges, 3), "d_filleted_2deg"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.min_wall_draft.draft_deg", 2.0, 0.05)], "drafted box, R3 top fillets: min draft stays 2"


@fixture("d_mixed_direction", "draft")
def _():
    a = Pos(-25, 0, 0) * drafted_box(deg=2)
    b = Pos(25, 0, 20) * (Rot(180, 0, 0) * drafted_box(deg=2))
    return stl(a + b, "d_mixed_direction"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.drafted_wall_mean_draft_deg", 2.0, 0.02)], "two bodies' walls open in OPPOSITE directions; does the report say so anywhere but per-face?"


@fixture("d_tilted_20deg_pull_z", "draft")
def _():
    return stl(Rot(20, 0, 0) * drafted_box(deg=2), "d_tilted"), [("draft.drafted_wall_mean_draft_deg", 2.0, 0.05)], "part modelled 20 deg off-axis, agent uses --pull z: truth is 2 deg about the part's own axis"


@fixture("d_tilted_20deg_pull_vec", "draft", pull="0,-0.3420201,0.9396926")
def _():
    return stl(Rot(20, 0, 0) * drafted_box(deg=2), "d_tilted"), [("draft.zero_draft_wall_area_mm2", 0, 0), ("draft.drafted_wall_mean_draft_deg", 2.0, 0.02)], "same part, pull given as the right vector"


# ---------------------------------------------------------------- UNDERCUTS
@fixture("u_cup_clean", "undercut")
def _():
    return stl(cup(), "u_cup"), [("undercuts.candidate_area_mm2", 0, 0)], "open drafted cup: no undercut"


@fixture("u_cup_side_hole", "undercut")
def _():
    part = cup() - Pos(19, 0, 9) * (Rot(0, 90, 0) * Cylinder(3, 8))
    return stl(part, "u_cup_side_hole"), [("undercuts.candidate_area_mm2", math.pi * 6 * 2.0, 0.2)], "D6 hole through a 2 mm side wall: bore = pi*6*2"


@fixture("u_snap_hook", "undercut")
def _():
    return stl(hook(), "u_snap_hook"), [("undercuts.candidate_area_mm2", 30 + 30 + 80, 0.1)], "lip underside 30 + base under lip 30 + post wall under lip 80 (trapped)"


@fixture("u_snap_hook_window", "undercut")
def _():
    return stl(hook(window=True), "u_snap_hook_window"), [("undercuts.candidate_area_mm2", 0, 0)], "same hook with a shut-off window below the lip: moldable, expect 0"


@fixture("u_groove", "undercut")
def _():
    part = Pos(0, 0, 15) * Cylinder(15, 30) - Pos(0, 0, 15) * (Cylinder(15.1, 4) - Cylinder(13, 4))
    truth = 2 * math.pi * (15 ** 2 - 13 ** 2) + 2 * math.pi * 13 * 4
    return stl(part, "u_groove"), [("undercuts.candidate_area_mm2", truth, 0.05)], "annular groove on a cylinder pulled along its axis"


@fixture("u_hourglass", "undercut")
def _():
    part = revolve(Plane.XZ * Polygon((0, -15), (12, -15), (5, 0), (12, 15), (0, 15), align=None), Axis.Z)
    truth = 2 * math.pi * (12 + 5) * math.hypot(15, 7)
    return stl(part, "u_hourglass"), [("undercuts.candidate_area_mm2", truth, 0.05)], "hourglass: both cones shadow each other"


@fixture("u_holes_along_pull", "undercut")
def _():
    plate = Pos(0, 0, 5) * Box(60, 40, 10)
    part = plate - Pos(-20, 0, 5) * Cylinder(4, 10) - Pos(0, 0, 8) * Cylinder(4, 4) - Pos(20, 0, 2) * Cylinder(4, 4)
    return stl(part, "u_holes_along_pull"), [("undercuts.candidate_area_mm2", 0, 0)], "through hole, blind from top, blind from bottom: none is an undercut"


@fixture("u_shadow_coarse", "undercut")
def _():
    return stl(shadow(), "u_shadow_coarse"), [("undercuts.candidate_area_mm2", 3000 + 3000 + 2000, 0.05)], "big shelf over a plate, CAD-coarse mesh (one ray per triangle centre)"


@fixture("u_shadow_fine", "undercut")
def _():
    mesh = trimesh.load(stl(shadow(), "u_shadow_coarse"), force="mesh")
    for _ in range(4):
        mesh = mesh.subdivide()
    return tm(mesh, "u_shadow_fine"), [("undercuts.candidate_area_mm2", 8000, 0.05)], "same part, subdivided 4x"


@fixture("u_dovetail_pull_z", "undercut")
def _():
    block = Pos(0, 0, 10) * Box(40, 30, 20)
    slot = extrude(Plane.XZ * Polygon((-4, 20), (4, 20), (7, 14), (-7, 14), align=None), 20, both=True)
    truth = 2 * math.hypot(3, 6) * 30 + 2 * 3 * 30
    return stl(block - slot, "u_dovetail"), [("undercuts.candidate_area_mm2", truth, 0.05)], "dovetail slot: 2 slanted walls + the floor strips under them"


@fixture("u_dovetail_pull_y", "undercut", pull="y")
def _():
    return FIX / "u_dovetail.stl", [("undercuts.candidate_area_mm2", 0, 0)], "same slot pulled along its length: free"


@fixture("u_internal_void", "undercut")
def _():
    return stl(Box(40, 40, 40) - Sphere(10), "u_internal_void"), [("undercuts.candidate_area_mm2", 4 * math.pi * 100, 0.05)], "sealed void: unmoldable, all of it is a candidate"


@fixture("u_two_bodies", "undercut")
def _():
    a = trimesh.creation.box(extents=(40, 30, 5))
    b = trimesh.creation.box(extents=(40, 30, 5)); b.apply_translation((0, 0, 12.5))
    return tm(trimesh.util.concatenate([a, b]), "u_two_bodies"), [("undercuts.candidate_area_mm2", 0, 0)], "an assembly STL (lid over base): two separate parts, neither has an undercut"


@fixture("u_open_shell", "undercut")
def _():
    mesh = trimesh.load(stl(cup(), "u_cup"), force="mesh")
    keep = ~((mesh.face_normals[:, 2] > 0.99) & (mesh.triangles_center[:, 2] > 19.9))
    mesh.update_faces(keep)
    return tm(mesh, "u_open_shell"), [("mesh.watertight", False, None)], "cup with its roof deleted (broken export): does the report flag it beyond watertight=false?"


# ---------------------------------------------------------------- WALL THICKNESS (default --samples)
@fixture("t_plate_3mm", "thickness", samples=2000)
def _():
    mesh = trimesh.creation.box(extents=(60, 40, 3))
    for _ in range(3):
        mesh = mesh.subdivide()
    return tm(mesh, "t_plate_3mm"), [("wall_thickness.median_mm", 3.0, 0.03), ("wall_thickness.p05_mm", 3.0, 0.1)], "flat 3 mm plate"


@fixture("t_plate_3mm_12tris", "thickness", samples=2000)
def _():
    return tm(trimesh.creation.box(extents=(60, 40, 3)), "t_plate_12tris"), [("wall_thickness.samples_valid", 2000, 0.5)], "same plate as CAD exports it (12 triangles): how many samples?"


@fixture("t_cup_2mm", "thickness", samples=2000)
def _():
    return stl(cup(), "u_cup"), [("wall_thickness.median_mm", 2.0, 0.05), ("wall_thickness.p05_mm", 2.0, 0.15)], "2 mm shelled cup"


@fixture("t_cube_60", "thickness", samples=2000)
def _():
    mesh = trimesh.creation.box(extents=(60, 60, 60))
    for _ in range(4):
        mesh = mesh.subdivide()
    return tm(mesh, "t_cube_60"), [("wall_thickness.p05_mm", 60, 0.1), ("wall_thickness.median_mm", 60, 0.1)], "solid 60 mm cube: every section is 60 thick"


@fixture("t_rib_boss", "thickness", samples=2000)
def _():
    part = Pos(0, 0, 1.25) * Box(60, 40, 2.5) + Pos(0, 10, 7.5) * Box(40, 1.5, 10) + Pos(-15, -8, 7.5) * Cylinder(4, 10)
    return stl(part, "t_rib_boss"), [("wall_thickness.median_mm", 2.5, 0.1), ("wall_thickness.min_mm", 1.5, 0.1), ("wall_thickness.max_mm", 8.0, 0.6)], "2.5 plate, 1.5 rib, solid D8 boss (the sink risk): min 1.5, thick tail >= 8"


@fixture("t_wedge_1_to_5", "thickness", samples=2000)
def _():
    part = extrude(Plane.XZ * Polygon((0, 0), (60, 0), (60, 5), (0, 1), align=None), 15, both=True)
    mesh = trimesh.load(stl(part, "t_wedge"), force="mesh")
    for _ in range(3):
        mesh = mesh.subdivide()
    return tm(mesh, "t_wedge"), [("wall_thickness.min_mm", 1.0, 0.25), ("wall_thickness.max_mm", 5.0, 0.15)], "wedge 1 mm to 5 mm"


@fixture("t_hollow_cube", "thickness", samples=2000)
def _():
    return stl(Box(40, 40, 40) - Sphere(10), "u_internal_void"), [("wall_thickness.min_mm", 10.0, 0.1), ("wall_thickness.max_mm", 30.0, 0.5)], "cored cube: true min wall 10 (face to void)"


@fixture("t_tube_2mm", "thickness", samples=2000)
def _():
    return stl(Pos(0, 0, 20) * (Cylinder(10, 40) - Cylinder(8, 40)), "t_tube"), [("wall_thickness.median_mm", 2.0, 0.08)], "pipe, 2 mm wall"


# ---------------------------------------------------------------- PROJECTION / SCALE
@fixture("p_plate_hole", "projection")
def _():
    return stl(Pos(0, 0, 2) * Box(100, 100, 4) - Pos(0, 0, 2) * Cylinder(10, 4), "p_plate_hole"), [("projection.projected_area_mm2", 10000 - math.pi * 100, 0.005)], "plate minus D20 hole"


@fixture("p_cube_diagonal", "projection", pull="1,1,1")
def _():
    return tm(trimesh.creation.box(extents=(20, 20, 20)), "p_cube"), [("projection.projected_area_mm2", math.sqrt(3) * 400, 0.01)], "cube along its body diagonal: sqrt(3)*a^2"


@fixture("p_pin_d3", "projection")
def _():
    return stl(Pos(0, 0, 2.5) * Cylinder(1.5, 5), "p_pin", tol=0.005, ang=0.05), [("projection.projected_area_mm2", math.pi * 2.25, 0.03)], "D3 pin: 0.25 mm raster on a 3 mm feature"


@fixture("p_big_2m", "projection")
def _():
    return tm(trimesh.creation.box(extents=(2000, 1500, 100)), "p_big"), [("projection.projected_area_mm2", 3.0e6, 0.005)], "2 m panel: raster must coarsen"


@fixture("p_inches", "scale")
def _():
    return tm(trimesh.creation.box(extents=(4, 3, 2)), "p_inches"), [("scale.units_suspect", True, None)], "a 4 x 3 x 2 INCH part exported unitless"


@fixture("p_metres", "scale")
def _():
    return tm(trimesh.creation.box(extents=(0.1, 0.08, 0.02)), "p_metres"), [("scale.units_suspect", True, None), ("projection.projected_area_mm2", None, None)], "100 mm part exported in metres: projection must not print 0.0 as a fact"


def run(fx):
    path, checks, note = fx["build"]()
    cmd = ["/usr/bin/time", "-l", PY, str(TOOL), "measure", str(path), f"--pull={fx['pull']}", "--samples", str(fx["samples"])] + fx["extra"]
    proc = subprocess.run(cmd, capture_output=True, text=True, timeout=900)
    rss = re.search(r"(\d+)\s+maximum resident set size", proc.stderr)
    secs = re.search(r"([\d.]+) real", proc.stderr)
    try:
        report = json.loads(proc.stdout)
    except Exception as exc:  # noqa: BLE001
        report = {"_parse_error": f"{type(exc).__name__}: {exc}", "_stdout_head": proc.stdout[:200]}
    faces = get(report, "mesh.triangle_count")
    rows = []
    for dotted, truth, tol in checks:
        got = get(report, dotted)
        if truth is None:
            verdict = "INFO"
        elif isinstance(truth, bool) or tol is None:
            verdict = "ok" if got == truth else "MISS"
        elif got is None:
            verdict = "MISS(null)"
        elif truth == 0:
            verdict = "ok" if abs(got) <= 1e-6 else "MISS"
        else:
            verdict = "ok" if abs(got - truth) <= tol * abs(truth) else "MISS"
        rows.append((dotted, truth, got, verdict))
    return {"name": fx["name"], "family": fx["family"], "note": note, "faces": faces, "exit": proc.returncode,
            "rss_mb": round(int(rss.group(1)) / 1e6) if rss else None, "secs": float(secs.group(1)) if secs else None,
            "rows": rows, "report": report}


def main():
    wanted = sys.argv[1:]
    results = []
    for fx in FIXTURES:
        if wanted and not any(fx["name"].startswith(w) for w in wanted):
            continue
        res = run(fx)
        results.append(res)
        print(f"\n[{res['family']}] {res['name']}  faces={res['faces']} exit={res['exit']} {res['secs']}s rss={res['rss_mb']}MB\n    {res['note']}")
        for dotted, truth, got, verdict in res["rows"]:
            t = f"{truth:.2f}" if isinstance(truth, float) else str(truth)
            print(f"    {verdict:10s} {dotted}: truth={t} got={got}")
    (HERE / "battery_results.json").write_text(json.dumps(results, indent=1, default=str))


if __name__ == "__main__":
    main()
```

</details>

earthtojake:
Re-reviewed at dfeb6f8ff. The consolidation is right and the tool is in much better shape. I checked 47 things I asked for or that your description claims, and 45 hold.

Verified by running it: the wall-thickness family is gone with no dead code, no leftover JSON keys and no dangling references, and the `$dfam-check` pointer is accurate down to the advice about `max_mm`; `measure` marks a failed family partial, names it and exits 2, while a clean report exits 0 and an unloadable mesh exits 1; the build123d sphere fixture that used to emit bare `NaN` now parses as valid JSON; `--units` converts every reported length and area; `--pull -z` and `--pull=-z` both work. The tangent rule fixed the sphere that read 261 mm² coarse and 67 mm² fine, and the tray's minimum draft no longer moves between exports. Undercut resampling is the standout: the shelf that read 9,500 against a true 8,000 now lands, and so do the dovetail floor strips and the snap hook. One skill, three references, `reports/` gone, docs page and AGENTS.md updated.

Four things to fix. The first is my own advice applied more literally than I meant it, so that one is on me.

## Blockers

**`min_wall_draft` hides real zero-draft walls.** I asked for a minimum pooled over faces above an area floor, to stop the single-facet noise. `_bulk` implements it as a 95%-of-area prefix, which is a fraction test with no absolute floor in it, so a real wall that matters drops out of the number the skill tells the agent to quote.

A 2 degree drafted tray, 200 x 150 x 100, with one dead-vertical boss wall:

| Boss | True zero-draft wall | `min_wall_draft` | Same report's `zero_draft_wall_area_mm2` |
|---|---|---|---|
| 20 x 20 | 400 mm² | **2.0 deg** | 880 |
| 40 x 40 | 1600 mm² | **2.0 deg** | 3360 |
| 60 x 60 | 3600 mm² | 0.0 deg | 6919 |

At 40 x 40 the report contradicts itself in three places: `min_wall_draft` says 2.0, `zero_draft_wall_area_mm2` says 3360, and `lowest_draft_wall_faces[0]` says 0.0 over 1600 mm². The face is planar, axis-aligned and 16 triangles, so nothing about it is a sliver, and it clears the stated 129.77 mm² floor twelve times over. `references/injection-molding.md:57` tells the agent to quote `min_wall_draft`, so this is the figure that reaches a review: a part with 3360 mm² of undrafted wall reads as drafted throughout.

Keep the absolute floor and drop the coverage prefix, or report both numbers and let the agent see the disagreement.

**`pulls` swallows a failed undercut family.** Block `rtree` and run both verbs on the same mesh:

```
measure: exit 2, partial true,  partial_sections ['undercuts'], undercuts.error "No module named 'rtree'"
pulls:   exit 0, partial false, no partial_sections, candidate_area_mm2 [null, null, null], no error anywhere
```

`_pull_facts` calls `_safe(_undercut_facts, ...)` and then keeps only `candidate_area_mm2`, so the `error` key it returns is discarded. `_mark_partial` scans top-level report values, and `report['pulls']` carries no error for it to find. An injected `RuntimeError` reproduces the same split.

This is the silent "no findings" that the partial machinery exists to prevent, and your own header says so: "neither a JSON reader nor a shell caller can take a partial report for a complete one." `pulls` is the verb an agent runs first to choose a pull direction.

## The rest

**The tangent rule reaches past tessellation artifacts into real walls.** A short flat vertical wall between two fillets is reclassified as a curved tangent band and removed from `zero_draft_wall_area_mm2`. On a CAD-coarse export a real 2 mm zero-draft flange rim reads 0. The rule needs to ask whether the neighbouring facets sweep through a range of angles, not whether a wall happens to sit between curved faces.

**`min_wall_draft` is still mesh-dependent in one case,** despite the docs saying otherwise: where a fillet sweeps through zero draft it reads 0.308 deg on a coarse mesh and 2.000 deg on three finer ones. It also reads 1.823 deg on a true 3.000 deg conic wall.

**`body_count` counts shells,** so a single watertight solid with a sealed internal void is reported as "an assembly or a split export, not one molded part". A cored part is the standard molding fix for a thick section, so this fires on the geometry the skill exists to encourage.

**A pull vector whose first component is negative flips the whole vector.** `--pull=-1,0,1` measures along `(-1,0,-1)`. The tool then reports a different axis than the one asked for, with no complaint.

**Above 200k faces the undercut probe stops subdividing** and reports a probe edge 437 times the part size, with nothing in the report saying coverage was dropped.

**A NaN that reaches the report exits 1 with a Python traceback** instead of the JSON error the header documents.

**The example project is broken on this head.** `scripts/verify.py` writes into `reports/`, which this PR deletes and `.gitignore` does not recreate, so the README's own Reproduce steps raise `FileNotFoundError`. The README also still quotes cone-sampled wall-thickness figures and attributes them to a command that no longer measures thickness.

Smaller things: `_RASTER_CELL_BUDGET` is unreachable, so the declared memory cap never fires; the rasteriser allocates full-raster float64 temporaries per triangle, so a 12-triangle panel peaks 35 times higher than it needs to; argparse usage errors exit 2 and collide with your new "exit 2 means partial"; `units_suspect` still passes a 4 x 3 x 2 inch part and calls it a plausible molded-part envelope; `_pooled_faces` hardcodes `zero_tol=0.05`, so a run with `--zero-tol` set calls the same faces both zero-draft and drafted in one report.

On the follow-up: `dfam-check` folding into `dfm` is still ours to do, not yours. The thickness method it needs is the one we discussed, thin end from a single normal ray and thick tail from a cone, cast against the whole mesh rather than per body, and the fixtures in the battery I attached last time are its acceptance tests.

earthtojake:
Re-reviewed at 24f24d32c and pushed one commit rather than send you back another round.

Both blockers are gone and each survives its inverse: a 400 mm² dead-vertical boss sets `min_wall_draft` to 0.000 while a plain 2 degree drafted box still reads 2.000, so the area floor kept the job it was added for, and `pulls` exits 2 with `partial: true` naming the section. Also confirmed: the fillet sweeping through zero reads 2.000 at four mesh densities, a sealed void is one body with `internal_void_count: 1`, `--pull=-1,0,1` measures the vector asked for, a NaN exits with JSON and no traceback, and `verify.py` runs end to end into `tmp/`. The known-answer battery shows no regressions anywhere.

**d1bca7d32** fixes the one regression this delta introduced. A region is read as tangent when it has smooth leaning neighbours both ways and is no taller than they are. OCCT fans a small bore, so every facet spans the full depth and the region is exactly as tall as its bounds:

| | `zero_draft_wall_area_mm2` | `zero_draft_tangent_area_mm2` |
|---|---|---|
| dfeb6f8ff, the parent | 6232.8 | 0.0 |
| 24f24d32c | 6190.4 | 42.4 |
| with this commit | 6232.8 | 0.0 |

That 42.4 mm² is the tray's own four Ø4.5 through-bores, the most ordinary zero-draft feature there is, and it is why the README's 6,233 stopped reproducing.

Size alone cannot separate them, so I added the other half of what a tangent line is. It is a CROSSING: the surface passes through parallel, so the leaning neighbours sit either side of the region ALONG the pull. Around a fanned bore they sit beside it spanning the same depth and share a centre. Measured across the fixtures, a sphere's band separates by 0.89 to 1.67 of its own height and the tray's bores by 0.33, so the threshold is 0.6 with margin either side. A region must now cross as well as fit. Erring low was deliberate: a missed suppression reports an artifact, a wrong suppression hides a wall.

The new test uses the tessellation a CAD export actually produces, on a drafted plate so the bore is the only zero-draft face, and the existing sphere and uv-sphere suppression tests still pass unchanged. 51 pass.

Three I left for you, none blocking. `largest_zero_draft_faces` did not get the conversion: `_draft_facts` still pools with `_pooled_faces`, so a Ø20 cylinder's zero-draft wall reads as 8 chord entries of 44.84 mm², and that is the list an agent reads to find out where. `undercuts.method` prints one probe edge for a multi-body file, the last body's, contradicting `probe_resolution_note`. And `pulls` is still the memory hazard at 2.01 GB and 43.9 s on a 786k-face export, where `measure` on the same mesh is 1.65 GB.

On the conic wall: disclosing it with `spread_deg` and `reading_note` is the right call and I am not asking for more. One note for later, `test_a_conic_wall_reads_its_built_draft_at_any_usable_mesh` picks three tolerances and asserts the minimum stays above `built - 0.7`, and three of five ordinary export settings land under that floor, so the 1.823 reading on a true 3.000 would pass unnoticed.

PR #428, opened 2026-09-18
Desktop: engineering drawings (Sheet panel, sheet editing, send to chat)

amywork777:
Desktop port of #427 (engineering drawings), re-based onto the rebuilt `claude/desktop-app` as one commit.

## In Hardcore

- A dimensioned DXF opens as the printed sheet (ezdxf render via `/__cad/drawing`), in plan view, on paper.
- The side panel for a drawing is a **Sheet** panel instead of Material/Bends/Layers: paper, scale, units, projection, revision and sheet n of m read from the title block; dimension units (as drawn / mm / in), precision and text size; line weight; switches for hidden lines, centre lines, dimensions, notes, title block.
- **Editing from the sheet**: click a dimension (red on the sheet) and give it a tolerance or fit; *Add dimension* picks two points; *Move views* drags a view. Every edit is previewed by the server and never written to the DXF. **Send to chat** delivers the exact drawing-API change through `host.promptContext` (the same channel as references and captures), so the agent edits the script and reruns it. The script stays the only source; no links between generated files.
- Re-renders swap the sheet image in place (no flash).

## Shared with main (same code as #427)

`cadgen.drawing` + `skills/drawing` (tolerances, hole callouts in standard symbols, general tolerance note, revision table, XDATA view tags with placement maps), the DXF parser in `packages/core`, the SVG route with dimension restyling and preview edits, and their tests.

## Notes for review

- The only merge conflict was the opening frame in `CadViewer.js`: kept the new `zoomRuntimeToBounds` fit and pass the plan-view direction/up when a sheet opens.
- Verified in the built app on the clevis demo drawing: tolerance, new dimension, moved view, and the request landing in the composer.
- `packages/core` and `packages/ui` suites pass apart from the listen-EPERM tests that need a socket; Python drawing suites pass.


amywork777:
Closing: your point 5 applies here, and the viewer half is all this PR is. #427 now carries the whole feature as `cadgen.eng_drawing` writing one PDF, with no Sheet panel, no sheet editing, no `/__cad/drawing`, no `drawing_svg.py` and no CADGEN XDATA on either side.

The desktop port of the DXF **rendering** (DIMENSION blocks, arrowheads, leaders, text alignment) is not lost — it is #430, off `main`, without the tab swap. If that lands, the desktop shell picks it up through `packages/core` with no separate port.

PR #377, opened 2026-09-10
feat(plugin): add QwenPaw plugin with CAD, robotics and fabrication s…

CallMeHFK:
Bundle the repo's skills (cad, cad-viewer, srdf, urdf, step-parts, bambu-labs, ...) as a QwenPaw plugin under .qwenpaw-plugin/, including plugin.json manifest, backend entry (plugin.py), per-skill SKILL.md/references/scripts, i18n descriptions, and a global test suite (tests/python/global/test_qwenpaw_plugin.py).

Also update AGENTS.md, CONTRIBUTING.md, README.md, docs landing page and sync-version script for the new plugin surface.

PR #429, opened 2026-09-21
Ship the skill library to QwenPaw, ZCode, and Cursor installs

CallMeHFK:
The repository root is already the plugin package for Claude Code, Codex, and the Skills CLI. Three loaders could not use it as-is: QwenPaw installs a plugin by copying one directory, so its skills have to travel inside it; ZCode reads a manifest of its own; Cursor discovers a plugin only through `.cursor-plugin/`. This adds those three surfaces. No skill changes, no `packages/`, no Viewer.

- **QwenPaw** (`.qwenpaw-plugin/`): `plugin.json` and `plugin.py` register the bundled `skills/` copy as a skill provider, add a `/cad-setup` preflight that reports the `cadgen` runtime (missing distribution, each skill's pin through `cadgen doctor`, stray viewer and daemon processes), and provision `bambu-labs` and `sendcutsend` disabled, because those two reach real machines. `skills/` under it is generated, and `test_qwenpaw_plugin.py` byte-compares it against the canonical tree in both directions.
- **ZCode** (`.zcode-plugin/`): a manifest and a marketplace entry sourced from the repository root, plus the `~/.zcode/skills` destination in `scripts/install/install-skills.sh` and its uninstaller.
- **Cursor** (`.cursor-plugin/plugin.json`): a repo-root manifest pointing at `./skills/`, the same shape Codex and ZCode use. The local install has to be cloned *into* `~/.cursor/plugins/local` -- Cursor loads a symlink there only when its target resolves inside that folder, so linking an existing checkout installs nothing at all.
- **Wiring shared by all three**: `scripts/release/sync-version.mjs` stamps every manifest from `VERSION`; the `infra` path filter in `.github/workflows/test.yml` names `.cursor-plugin/**`; `check-builds.sh` extends its LFS scope to the generated copy (defence in depth -- the lockstep test already fails a divergent copy); `test_plugin_manifests.py` covers each provider manifest and `test_skill_catalog_sync.py` binds the hand-written skill catalogs to `skills/`.

18 files carry the change, about 1.1k lines; the other 89 are the generated `.qwenpaw-plugin/skills/` copy. `VERSION` is unchanged, and the release workflow stamps the pins separately.

Supersedes #377: that branch predates the 0.6.x line (its `VERSION` was 0.5.1) and its bundled copy sat 43 files behind `skills/`, so it is rebuilt here from current `main` rather than updated in place.

Validation: 147 policy tests via `scripts/test/test-global.sh` (one expected legacy-layout skip); `scripts/release/check-version.sh`; `scripts/github-workflows/check-builds.sh`; `npm --prefix apps/docs run check`. Not run locally: the Viewer client's JS suite, because `apps/viewer/node_modules` is absent in this checkout and `react` will not resolve -- no Viewer or `cadgen-js` source is in this diff, and #424 shows that job passing upstream for an `infra`-only change, which is the class this diff falls in.


CallMeHFK:
@earthtojake ready for review when you have a moment — could you approve the workflow run? As a fork PR the `Test` run is sitting at `action_required`, so none of the required checks have executed yet.

Three plugin loaders could not use the repository-root package; this adds them. No skill changes, nothing under `packages/` or the Viewer, and `VERSION` is untouched.

- **QwenPaw** (`.qwenpaw-plugin/`) — installs by copying one directory, so it carries a generated `skills/` copy; `plugin.py` registers it as a skill provider, adds a `/cad-setup` runtime preflight, and provisions `bambu-labs` / `sendcutsend` disabled because those reach real machines.
- **ZCode** (`.zcode-plugin/`) — manifest plus marketplace entry sourced from the repo root, and the `~/.zcode/skills` destination in the link installer.
- **Cursor** (`.cursor-plugin/plugin.json`) — repo-root manifest pointing at `./skills/`.

Reviewing: 18 files carry the change (~1.1k lines); the other 89 are the generated `.qwenpaw-plugin/skills/` copy, byte-locked to `skills/` by `test_qwenpaw_plugin.py`. History is 3 linear commits, one per loader.

Validated locally: `scripts/test/test-global.sh` (147 tests, one expected legacy-layout skip), `scripts/release/check-version.sh`, `scripts/github-workflows/check-builds.sh`, `npm --prefix apps/docs run check`. The Viewer client's JS suite did not run here — `apps/viewer/node_modules` is absent in this checkout — and no Viewer or `cadgen-js` source is in this diff; #424 shows that job passing for an `infra`-only change, which is the class this diff falls in, so CI will cover it.

Two documented claims were corrected against primary sources while building this, both worth a second pair of eyes: Cursor loads a symlink in `~/.cursor/plugins/local` only when its target resolves inside that folder (so the install clones into place — a linked checkout installs nothing), and the QwenPaw loader enforces `qwenpaw_version`'s max as well as its min, which makes the `2.99.0` ceiling load-bearing rather than decorative.

Supersedes #377, which predated the 0.6.x line.

earthtojake:
@CallMeHFK we need to avoid duplicating all the skills inside these plugins: do you know if these plugins support resolving symlinks on install? This is what we do for codex, grok and claude plugins.

CallMeHFK:
Checked this against QwenPaw's source at `origin/main` (`c11c9d7c`) rather than from memory. Short answer: yes on one of its two install paths, no on the other.

First, scope: only `.qwenpaw-plugin/skills/` duplicates. The ZCode and Cursor manifests are `"skills": "./skills/"` pointers at the repo root, the same mechanism the Claude, Codex and Grok surfaces already use, so nothing is duplicated there.

**Directory install dereferences symlinks -- measured.** `plugins/loader.py:1286` is `shutil.copytree(source_path, stage_dir)` with no `symlinks=` argument, so the default (`symlinks=False`) applies, which follows links and copies their targets. A sandbox with `.qwenpaw-plugin/skills -> ../skills` installed as a real `skills/cad/SKILL.md` with the file body intact. The CodeQL `realpath` guard in `resolved_plugin_manifest_path` only covers `plugin.json`, so it does not object. That is the 89-file copy and ~10.4k lines of this diff gone.

**ZIP install cannot carry the link at all.** `app/routers/plugins.py` `_safe_extract_zip` rejects Zip Slip by member name and then calls `zip_ref.extractall`, and stdlib extraction does not recreate symlink entries -- the member lands as a *regular file whose contents are the text `../skills`*. `_resolve_skills_dir()` then fails its `is_dir()` probe and the plugin logs "no skills directory found ... skill registration skipped": a plugin with zero skills installs successfully and no error. Zipping the repository root instead of the plugin directory does not help, since extraction never restores links regardless of what else is in the archive. `cli/plugin_commands.py:install` accepts a local directory or a local `.zip`, so this is a supported route, not a hypothetical one.

So the symlink buys the deduplication you want only if the plugin ships as a directory install. Three costs I would not absorb without your call:

1. `.qwenpaw-plugin/README.md` documents the ZIP route (and `qwenpaw plugin install <dir>` is the only one that survives) -- it would have to go, or silently ship an empty plugin to anyone who follows it.
2. `check-builds.sh check_tree_has_no_symlinks()` rejects any tracked symlink repo-wide, which is the law AGENTS.md says not to relax because Codex drops symlinks silently and the Skills CLI dereferences them. A link confined to `.qwenpaw-plugin/` is inert for the other loaders (none of them read that directory), but the gate as written fails it regardless, so this needs a scoped exception rather than a workaround.
3. A tracked relative symlink is hostile to the Windows path this repo cares about: a checkout with `core.symlinks=false` materialises it as a text file. The lockstep test would catch that loudly rather than silently, but it is a trap for contributors on that platform.

Recommendation, and tell me which you want: switch `.qwenpaw-plugin/skills` to a tracked symlink, drop the committed copy and the ZIP instructions, and add the narrow gate exception (one path, `.qwenpaw-plugin/skills`, with the reason recorded beside the Codex/Skills-CLI note); or keep the committed copy as the price of ZIP installability. I have not changed anything yet because option one edits a check whose file says do not relax it.

CallMeHFK:
Answering the symlink question, then showing the change that removes the duplication without one: `86396fad` drops the copy -- 120 files to 18, +13,249 to +1,199.

**Symlinks are not something an install can turn on.** There is no config key, env var or install flag for link handling: every `shutil.copytree` call site in QwenPaw uses the default (`symlinks=False`) and exposes no parameter, and the plugin install path's only knob is `replace_files` (`--force`). A plugin cannot affect it either -- `plugin.py` is the thing being copied, and it runs after the copy. Checked against `agentscope-ai/QwenPaw` at `c11c9d7c`.

It gets worse before it gets better: links in skill content are a rejected input class, not a disabled feature. `workspace_service.py` refuses a linked source, scans the whole tree for links, refuses a linked install root, and `store.py` rejects symlink members in a skill ZIP outright. So the directory install only "working" with a link is `copytree` flattening it before any guard sees it, and the ZIP route (`_safe_extract_zip`) has no link check at all -- it leaves a text file containing `../skills`, the plugin resolves nothing, and registers zero skills behind one log line. I would not build the packaging on that.

**What landed instead.** The entry point resolves a `skills/` tree at load time: `plugins.cad.skills_dir` from QwenPaw's config, then a generated copy, then the checkout sibling -- and `/cad-setup` prints which won, since "pointing at a stale tree" is invisible from the workspace skill page. Nothing is needed from QwenPaw for this: `app/_app.py` builds `plugin_configs` from `config.plugins` and passes it to `load_all_plugins`, which hands it to `PluginApi` as `self.config`, and `register_skill_provider` does `Path(skills_dir)` with no requirement that the tree live inside the installed package. A configured path that fails to resolve stops resolution rather than falling through, so a moved checkout surfaces instead of being papered over by an old copy.

Two costs, documented in `.qwenpaw-plugin/README.md` rather than hidden: the install route does not pass config, so a hot-install while the app runs picks the tree up on the next start; and a ZIP of the plugin directory carries no skills, so that route needs the rsync step -- which is why the recipe stays as a command instead of as committed files.

One more option worth your eye, since it is QwenPaw's own mechanism: `config.skill_paths` is a list of additional read-only skill-pool roots with `~` expansion. Pointing it at this repo's `skills/` gives all thirteen skills with no plugin copy and no per-plugin config at all -- the user downloads them per workspace. What is lost is the parts the plugin exists for: default-enable on provision, the `bambu-labs`/`sendcutsend` gate, and `/cad-setup`. If you would rather have that split -- `skill_paths` for the library, plugin for the runtime checks -- say so and I will reshape it that way.

Nothing here is a reply to your review being wrong: the copy was the simplest thing that worked and you asked whether it was necessary, and it is not. If you prefer the copy anyway, `86396fad` reverts cleanly on its own.

CallMeHFK:
Verified the config-pointer design against the real runtime rather than leaving it as a reading of their source. Harness: the installed QwenPaw 2.2.1's own `PluginLoader` over a copy of `.qwenpaw-plugin/` in a temp plugin dir, one process per scenario so no global registry leaks between cases.

| scenario | `api.config` | resolved |
| --- | --- | --- |
| config at a real checkout tree | `skills_dir=<clone>/skills` | `configured` |
| no config, sibling present | none | `checkout sibling` |
| config at a moved-away path | `skills_dir=/nope/moved-away/skills` | `configured path invalid`, no fallback |
| config blank | `skills_dir="   "` | falls through → `nothing reachable` |
| generated copy + sibling both present | none | `bundled copy` wins |

`discovered plugin ids: ['cad']` and `enabled=True` in every case, so `qwenpaw_version` `{min: 2.1.1, max: 2.99.0}` passes on a 2.2.1 runtime and a missing skills tree still leaves the plugin loadable -- it skips skills and keeps `/cad-setup`, which is the degradation the entry point documents.

The ZIP route, through their `_safe_extract_zip` and `_find_plugin_dir` rather than my paraphrase: a ZIP built exactly as `.qwenpaw-plugin/README.md` says (rsync recipe, then `zip -r`) extracts and resolves `bundled copy` with all 13 skills present; a ZIP without that step resolves nothing, which is the reason the README carries the warning instead of offering ZIP as an equal path.

One defect this surfaced and `67e1d116` fixes: the state reported the basename of a bad configured path, so `/nope/moved-away/skills` came out as `skills` -- the string a user needs to see was the one dropped. Now the full path, with an assertion pinning it.

Not covered here: an actual install into a live `~/.qwenpaw` (it would overwrite an existing plugin and need a restart), so workspace provisioning, the fabrication gate, and how `/cad-setup` renders remain unexercised end to end. What this change touches -- resolution and its failure modes -- is what the table above covers.

